PR #1, opened 2017-08-15
Create PubSub layer, and accompanying demo

huderlem:
PubSub layer includes the ISubscriber and IPublisher interfaces, along with the respective NetMQ implementations.

When a subscriber subscribes to a topic, it also provided a message handler callback, which processes the incoming message.  This allows any TPP modules to easily pick and choose which messages they care about and how to process them.  

The two demo projects, TestServer and TestClient, show the PubSub layer in action.  Additionally, they demonstrate how we can use dependency injection to ensure how the implementations of these services can be completely hidden from a TPP module, (and easily mock-able for functional/unit tests).


huderlem:
Latest commit adds full-fledged strong-typed PubSub messages.  The test server and client have been updated accordingly.

PR #7, opened 2017-08-20
First pass of config settings support

huderlem:
1. Add strongly-typed config classes.  These are instantiated using the `ConfigBase.Read<T>()` method.  When instantiated, they call into the `YamlConfigReader` to deserialize a list of config files in hierarchical order.  This means that each TPP service can augment and override settings in the core .yaml file with their own service-specific .yaml files.  Currently, it assumes that the .yaml files live in the same directory as the currently-executing assembly.

This pull request **does not** address on-the-fly config value reloading, or database-controlled config values because that needs more discussion.

2. Add new `TPPService` class which is meant to be the base class for all TPP services.  It controls the startup sequence, config files, and may do more interesting things in the future.  The existing Logging Service and test services were converted to inherit from it.


huderlem:
Added more strict rules around config files and values, meaning an exception will be thrown if any missing or unrecognized configs are detected.  Also added commandline support for specifying specific config values and config files.

PR #21, opened 2017-11-07
Add mock pub sub and chat service stub code

chfoo:
I was thinking that the chat service should be in two parts internally. `Chat.Client` is intended for code that touches the network so they could be mocked.`Chat.Service` is intended for code that interacts with pub sub. I'm not sure about the design of `TPPService` because it encapsulates everything into a `Run` method.


chfoo:
Closing because code is going off in a wrong direction

PR #22, opened 2018-04-18
Added a chat logger microservice

red031000:


src/TPPCore.Service.Common/TPPCore.Service.Common.csproj
```diff
@@ -8,6 +8,7 @@
     <PackageReference Include="CommandLineParser" Version="2.1.1-beta" />
     <PackageReference Include="log4net" Version="2.0.8" />
     <PackageReference Include="Microsoft.AspNetCore" Version="2.0.1" />
+    <PackageReference Include="RingBuffer" Version="1.5.1" />
```
chfoo: Any reason why RingBuffer is included in Common? I can only remember using this in Irc.Tests.

src/TPPCore.Service.ChatLogger/ChatLoggerService.cs
```diff
@@ -1,30 +1,34 @@
...
 
-			context.PubSubClient.Subscribe(ChatLoggerTopics.Raw,
-				(topic, message) =>
-				{
-					LogManager.LogMessage(message);
-				});
-		}
+        public void Run()
+        {
+            while (running) ;
```
chfoo: This works but you should use a synchronization primitive to put the thread to sleep instead of busy waiting. You could use `ManualResetEvent` to wait here and then in the `shutdown` method, call `set` to wake it back up.

PR #23, opened 2018-04-20
moved chatproviders and datamodels into own project

red031000:


chfoo:
@Felk We were talking about it and the idea resulted was that we didn't want to lump together things that look and behave similarly but had no dependency on each other.

aaaaaa123456789:
why am I _still_ getting emails for this pull

PR #24, opened 2018-05-06
added restful clients to chat and parrot and did some refactoring

red031000:


test/TPPCore.Service.Chat.Tests/DummyRestfulTest.cs
```diff
@@ -43,70 +39,45 @@ private ServiceTestRunner newServiceRunner()
...
-
-            Assert.Equal(HttpStatusCode.OK, result.Response.StatusCode);
-
-            var jsonDoc = result.JsonDoc;
-            var users = jsonDoc.Value<JArray>("users");
-
-            Assert.Equal(2, users.Count);
-            Assert.Equal("dummy", (string) users.SelectToken("[0].username"));
+            ChatClient client = new ChatClient(runner.Runner.Context.RestfulServer.Context.GetUri().ToString(), "dummyTest", httpClient);
+            var result = await client.GetRoomList("%23somechannel");
```
chfoo: It should be `#somechannel` as in IRC so the user doesn't have to know they need to percent encode the parameters.

src/TPPCore.ChatProviders/IRestEvent.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Newtonsoft.Json.Linq;
+
+namespace TPPCore.ChatProviders
+{
+    /// <summary>
+    /// Represents an object that is used for Rest API queries.
+    /// </summary>
+    public interface IRestEvent
+    {
+        JObject ToJObject();
```
chfoo: I thought we don't want to do this JObject pattern anymore.

src/TPPCore.ChatProviders/Providers/Twitch/TwitchProvider.cs
```diff
@@ -85,16 +85,58 @@ public string GetUsername()
...
+                users.Add(user);
+            }
+            foreach (string staffmember in staff)
+            {
+                var user = new ChatUser() { Username = staffmember };
+                Staff.Add(user);
+                users.Add(user);
+            }
+
+            var processedUsers = users.Distinct(new ChatUserEqualityComparer()).ToList();
```
chfoo: I think there might be some confusion about what the code is supposed to be doing here. 

Twitch IRC's room events are delayed or can be unavailable. What the code was doing was merging the IRC room list with a fresh call to the Twitch API. Because there is going to be duplicates, the Distinct method was supposed to make the room list contain unique `ChatUser`s. With the new changes, there isn't any deduplication besides the number of users in the room.

.editorconfig
```diff
@@ -1,7 +1,7 @@
 root = true
 
 [*]
-end_of_line = lf
+end_of_line = crlf
```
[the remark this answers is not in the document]
red031000: git automatically changes line endings to crlf, so it's better to have crlf rather than lf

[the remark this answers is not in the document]
chfoo: Maybe this config option should deleted and we can [add a .gitattributes file instead](https://help.github.com/articles/dealing-with-line-endings/)

src/TPPCore.Service.Common/AspNetUtils/HttpContextExtensions.cs
```diff
@@ -8,23 +8,22 @@ namespace TPPCore.Service.Common.AspNetUtils
     public static class HttpContextExtensions
     {
         /// <summary>
-        /// Sets the appropriate HTTP headers and writes the given JSON object.
+        /// Sets the appropriate HTTP headers and writes the given string object.
         /// </summary>
-        public async static Task RespondJsonAsync(this HttpContext context, JObject doc)
+        public async static Task RespondStringAsync(this HttpContext context, string doc)
```
chfoo: This confused me for a bit. The documentation could say that it writes a "JSON-formatted string" and not just any arbitrary string since it sets the content type to be JSON.

src/TPPCore.Client.Common/CommonClient.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static async Task PostAsync(Uri requestUri, string message, RestfulClient HttpClient)
+        {
+            HttpResponseMessage msg = await HttpClient.PostAsync(requestUri, new StringContent(message));
+            if (!msg.IsSuccessStatusCode)
+                throw new HttpRequestException();
+        }
+
+        public static string Escape(string toEscape)
+        {
+            return Regex.Replace(toEscape, @"[^\w%]", m => "%" + ((int)m.Value[0]).ToString("X2"));
```
chfoo: Why not use `Uri.EscapeDataString` instead? It handles recognizing reserved characters and also Unicode strings.

src/TPPCore.Client.Chat/TPPCore.Client.Chat.csproj
```diff
@@ -0,0 +1,18 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\TPPCore.ChatProviders\TPPCore.ChatProviders.csproj" />
+    <ProjectReference Include="..\TPPCore.Client.Common\TPPCore.Client.Common.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json">
+      <HintPath>C:\Program Files\dotnet\sdk\NuGetFallbackFolder\newtonsoft.json\9.0.1\lib\netstandard1.0\Newtonsoft.Json.dll</HintPath>
```
chfoo: Looks like this was added by accident?

PR #25, opened 2018-05-10
added timeout and ban to chatclient

red031000:


src/TPPCore.ChatProviders/Providers/Twitch/TwitchProvider.cs
```diff
@@ -59,7 +59,17 @@ public string GetUsername()
 
         public async Task SendPrivateMessage(string user, string message)
         {
-            await SendMessage("#jtv", $".w {user} message");
+            await SendMessage("#jtv", $".w {user} {message}");
```
chfoo: Good catch

PR #26, opened 2018-05-12
close #16 by implementing an emote class to be sent with ChatMessages

### LINKED ISSUE #16: Chatservice: Parse emotes
This feature depends on #8.

The chat service should recognize emotes supplied by the chat provider.
I suggest attaching a list of `EmoteOccurrence` for each message, where each occurrence consists of the start- and end-index in the message, information about the emote itself, and an image URL for other components to insert the emote images.

While initially only implented for Twitch Emotes, the approach should allow for potential other chat providers too.

red031000:


src/TPPCore.ChatProviders/Providers/Twitch/TwitchIrcProvider.cs
```diff
@@ -23,25 +26,105 @@ public TwitchIrcProvider()
...
+
+            if (value != "" && value != null)
+            {
+                List<EmoteOccurance> emoteOccurances = new List<EmoteOccurance> { };
+                Dictionary<string, Tuple<int, string>> keyValuePairs = new Dictionary<string, Tuple<int, string>> { };
+                List<string> emotes = value.Split('/').ToList(); // / seperates emote types
+                foreach (string emote in emotes)
+                {
+                    int.TryParse(emote.Split(':')[0], out int emoteId); // : seperates the ID from the indexes
+                    string indexes = emote.Split(':')[1];
```
chfoo: This is ok since we don't expect Twitch to throw garbage that bad at us, but it might be a good idea to only split once, check bounds of the result, and check the result of the TryParse.

Something like 
```cs
var splitEmoteColon = emote.Split(':');
if (splitEmoteColon.Length != 2 || !int.TryParse(splitEmoteColon[0], out int emoteId)
{
    continue;
}

File: src/TPPCore.ChatProviders/Providers/Twitch/TwitchIrcProvider.cs [hunk 2]
```diff
@@ -23,25 +26,105 @@ public TwitchIrcProvider()
...
+                    if (emote.Contains(',')) // , seperates emote occurances
```
chfoo: I'm not sure why this is needed. If we do `"0-10".Split(',')` it will still return an array but it will be length 1 with the original string as the item: ["0-10"]. This lets us use the foreach statement regardless of how many emotes there are.

src/TPPCore.ChatProviders/Providers/Twitch/TwitchIrcProvider.cs
```diff
@@ -23,25 +26,105 @@ public TwitchIrcProvider()
...
+                        List<string> occurances = indexes.Split(',').ToList();
+                        foreach (string occurance in occurances)
+                        {
+                            int.TryParse(occurance.Split('-')[0], out int firstIndex); // - seperates the start index from the end index
+                            int.TryParse(occurance.Split('-')[1], out int secondIndex);
+                            string name = string.Empty;
+                            for (int i = firstIndex; i <= secondIndex; i++)
+                            {
+                                name += chatMessage.TextContent[i];
+                            }
```
chfoo: I think you want to use `string.Substring(index, length)`?

Checking the result of `int.TryParse` and skipping this occurrence if it fails might be a good idea too but not necessary.

src/TPPCore.ChatProviders/Providers/Twitch/TwitchIrcProvider.cs
```diff
@@ -52,12 +135,74 @@ private Task userNoticeEventHandler(IrcClient client, Message message)
...
+            }
+            else
+            {
+                chatEvent.Emote = new LoyaltyEvent.Emotes
+                {
+                    Data = new Dictionary<string, Tuple<int, string>> { },
+                    Ranges = new List<EmoteOccurance> { }
+                };
+                Debug.Assert(chatEvent.Emote != null);
+            }
```
chfoo: I think putting this into a function instead of copy and pasting might be better. 

(This is something you might do in a underhanded code contest to sneak code into things Kappa)

src/TPPCore.ChatProviders/Providers/Twitch/TwitchIrcProvider.cs
```diff
@@ -52,12 +135,74 @@ private Task userNoticeEventHandler(IrcClient client, Message message)
...
             context.PublishChatEvent(chatEvent);
+
             return Task.CompletedTask;
         }
 
-        private Task whisperEventHandler(IrcClient client, Message message)
+        private Task WhisperEventHandler(IrcClient client, Message message)
```
chfoo: Don't forget about the whispers which also have tags

PR #28, opened 2018-05-15
implemented getuserid for twitchprovider

red031000:


src/TPPCore.ChatProviders/Providers/Twitch/TwitchProvider.cs
```diff
@@ -43,7 +46,20 @@ public void Shutdown()
...
         public string GetUserId()
         {
-            throw new System.NotImplementedException();
+            string clientID = context.Service.ConfigReader.GetCheckedValue<string>(
+                    "chat", "clients", ClientName, "client_id");
+            string url = NewApiUrl + "/users?login=" + GetUsername();
+            httpClient.DefaultRequestHeaders.Add("client-id", clientID);
+
+            HttpResponseMessage response = httpClient.GetAsync(url).Result;
+            string JsonString = response.Content.ReadAsStringAsync().Result;
```
chfoo: I think it would be best if `GetUserId` was changed to async through out since this call will be blocking the rest of the chat service.

PR #29, opened 2018-05-25
implemented database

red031000:


m4-used-rollout:
We need the emote service, so pulling

m4-used-rollout:
Pulling as soon as I get write access.

PR #41, opened 2020-04-02
add `PkmnSpecies` class for type-safe species handling

Felk:


Common/PkmnSpecies.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// </summary>
+    public sealed class PkmnSpecies : IComparable<PkmnSpecies>
+    {
+        private static readonly Dictionary<string, string> NoFlavors = new Dictionary<string, string>();
+
+        /// <summary>
+        /// Internal dictionary that keeps track of any registered pokedex data
+        /// by providing a lookup from species ids to species instances containing said pokedex data.
+        /// </summary>
+        private static readonly Dictionary<string, PkmnSpecies> Pokedex = new Dictionary<string, PkmnSpecies>();
```
red031000, line 25: would prefer some type of static constructor to register pokedex IDs automatically based on a json file or smth like that

Felk: I made it the way it is right now for multiple reasons:
- Use the species class _everywhere_, do not use a raw string _anywhere_ (except at application boundaries).
- Convenience of having the pokedex readily available in the species class.
- don't transitively make everything depend on Json.NET
- don't cause _everything_ to have a slow startup time due to parsing a large json file (which would include unit tests)

In my blob of local changes I actually do have a pokedex.json in the main executable project that gets loaded and registered at startup

Felk: I could put that blob of text in the xmldoc I guess

red031000: that's fair, though it does raise the question of how the pokedex will get registered

PR #42, opened 2020-04-06
Core executable project

Felk:


Felk:
m4 wanted to look at this as well. I'll request a review off of him

Core/Application.cs
```diff
@@ -0,0 +1,104 @@
...
+Options:
+  -h --help         Show this screen.
+  --config=<file>   Specifies the config file to use. [default: config.json]
+  --outfile=<file>  Specifies the file to output to. Default is printing to stdout.
+";
+
+        private static void Main(string[] args)
+        {
+            using var loggerFactory = LoggerFactory.Create(builder => builder
+                .AddFile("logs/tpp-{Date}.log")
```
m4-used-rollout, line 34: Logger file is not configurable? Maybe we could have it log only to console until it gets a log destination in the config

PR #43, opened 2020-04-11
add currency bank

Felk:


Persistence.MongoDB/README.md
```diff
@@ -0,0 +1,17 @@
+﻿This project implements the [Persistence interface](../Persistence) using a MongoDB database.
+It is designed to seamlessly function with the same database the old core uses.
+
+## run MongoDB
+
+- Install MongoDB 4.2 or higher. Get it [here](https://www.mongodb.com/download-center/community).
+- The server must be in [replica set mode](https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set/).
```
Felk, line 7: This is only required because transactions require the server to be in replica set mode. however, for development it might be just fine to not have multi-statement and multi-document transactions. However, this would require the code to detect this and do things differently. The connection string would be different, and it would branch at each transaction to just not do the transaction if the server is in standalone mode.

My gut feeling says a slightly more complicated one-time development system setup is the lesser evil here

Persistence.MongoDB.Tests/Repos/BankTest.cs
```diff
@@ -0,0 +1,209 @@
...
+        private IMongoDatabase _database = null!;
+        private IMongoCollection<TestUser> _usersCollection = null!;
+
+        private IBank<TestUser> _bank = null!;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _database = CreateTemporaryDatabase();
+            _bank = new Bank<TestUser>(
```
red031000, line 33: label parameters?

Persistence.MongoDB/Serializers/TransactionTypeSerializer.cs
```diff
@@ -0,0 +1,53 @@
...
+using MongoDB.Bson.Serialization;
+using Persistence.Models;
+
+namespace Persistence.MongoDB.Serializers
+{
+    public class TransactionTypeSerializer : EnumToStringUsingTranslationMappingSerializer<TransactionType>
+    {
+        public static readonly TransactionTypeSerializer Instance = new TransactionTypeSerializer();
+
+        private TransactionTypeSerializer() : base(new Dictionary<TransactionType, string>
```
red031000, line 12: I'm pretty sure there's more (or there should be anyway) if not, then that's probably an issue with our current core that we need to fix

Persistence/Models/TransactionLog.cs
```diff
@@ -0,0 +1,75 @@
...
+        Test, // whatever this was originally used for, but there are several thousand entries in the database.
+    }
+
+    /// <summary>
+    /// Transaction logs are read-only entities that get created by <see cref="IBank{T}"/> implementations.
+    /// They are purely for traceability and serve no functional purpose.
+    /// </summary>
+    // properties need setters for deserialization
+    // ReSharper disable AutoPropertyCanBeMadeGetOnly.Local
+    public class TransactionLog : IEquatable<TransactionLog>
```
red031000, line 39: operator==?

Persistence/Models/TransactionLog.cs
```diff
@@ -0,0 +1,75 @@
...
+        public bool Equals(TransactionLog? other) => Id == other?.Id;
+
+        public string UserId { get; private set; }
+        public int OldBalance { get; private set; }
+        public int NewBalance { get; private set; }
+        public int Change { get; private set; }
+
+        public DateTime CreatedAt { get; private set; }
+
+        public TransactionType Type { get; private set; } // TODO felk enum
```
red031000, line 51: todo?

Persistence/Repos/IBank.cs
```diff
@@ -0,0 +1,177 @@
...
+namespace Persistence.Repos
+{
+    /// <summary>
+    /// Exception thrown when a user related operation failed because the user did not exist.
+    /// </summary>
+    public class UserNotFoundException<T> : Exception
+    {
+        public T User { get; }
+
+        public UserNotFoundException(T user)
```
red031000, line 18: no message for the exception?

Persistence/Repos/IBank.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            User = user;
+        }
+    }
+
+    /// <summary>
+    /// A single atomic monetary transaction that can be performed.
+    /// </summary>
+    /// <typeparam name="T">User object type the transaction is performed on, typically <see cref="User"/></typeparam>
+    public readonly struct Transaction<T>
```
red031000, line 28: might it be an idea to have some inheritance for the types of users, rather than just passing in a T here

Persistence/Repos/IBank.cs
```diff
@@ -0,0 +1,177 @@
...
+            AdditionalData = additionalData ?? ImmutableDictionary<string, object?>.Empty;
+        }
+    }
+
+    /// <summary>
+    /// A bank is capable of safely handling some entity's numeric fields that represent a monetary value.
+    /// It offers functionality for safe transactional adjustments, money reserving, and database logging.
+    /// </summary>
+    /// <typeparam name="T">User object type this bank operates on, typically <see cref="User"/></typeparam>
+    public interface IBank<T>
```
red031000, line 65: see above

Persistence.MongoDB/Serializers/TransactionTypeSerializer.cs
Felk: i got this list by getting a fresh tokentransaction dump and doing a distinct(). Guess I forgot pokeyentransactions are a thing too. maybe this shouldn't be an enum after all, but a string

Persistence/Repos/IBank.cs
Felk: why? what's bad about being agnostic about what and what isn't a "user" as long as it meets the requirements?

red031000: well people can pass in objects that don't necessarily meet the requirements, even if it's just a constraint `where T : IUser` or something

Felk: I tried replacing the idFieldAccessor and currencyFieldAccessor parameters in the bank class with an interface:
```c#
public interface IBankUser
{
    public string Id { get; }
    public int Currency { get; protected internal set; }
}
```
that works, and makes the Bank implementation a lot easier, but then you cannot have banks on the same class but different currency fields anymore. That would require some sort of proxy classes like:
```c#
class PokeyenProxy : IBankUser
{
    private readonly User User;
    public PokeyenProxy(User user) => User = user;

    public string Id => User.Id;
    int IBankUser.Currency
    {
        get => User.Pokeyen;
        set => User.Pokeyen = value; // doesn't even compile because the setter is private
    }
}
```

Felk: imo not worth it

Felk:
I think I adressed everything

PR #44, opened 2020-05-25
ci: also use pull_request event

Felk:
this way the CI pipeline also executes the tests on a test merge (merge request with master) instead of just the source branch's HEAD

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -1,6 +1,6 @@
 name: .NET Core
 
-on: [push]
+on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


red031000:
this has the side affect of having 2 checks? or is this just because you're merging from a branch on master

Felk:
> 
> 
> this has the side affect of having 2 checks? or is this just because you're merging from a branch on master

2 checks for pull requests, yes. everything in a pull requests gets a check on a pre-merged state, and every pushed commit also gets checked by itself

red031000:
ah ok, makes sense

PR #48, opened 2020-05-31
add hex color parser

Felk:


ArgsParsing/Types/HexColor.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class HexColor
+    {
+        public string HexColorString { get; }
+
+        public HexColor(string hexColorString)
+        {
+            HexColorString = hexColorString;
+        }
+
+        public static implicit operator string(HexColor hexColor) => hexColor.HexColorString;
```
red031000, line 16: hashtag in front? idk if that's better than no hashtag though, but it makes it clear it's a hex colour

Felk: I don't know which is better either, but for _some_ guidance, I could make it follow TwitchLib's convention, which includes the hash

PR #54, opened 2020-06-23
rename BaseBank to more specific ReserveCheckersBank

Felk:


Persistence/Repos/IBank.cs
```diff
@@ -159,19 +159,14 @@ public async Task<int> GetReservedMoney(T user)
...
         public abstract Task<int> GetTotalMoney(T user);
 
-        public async Task<int> GetAvailableMoney(T user)
-        {
-            return await GetTotalMoney(user) - await GetReservedMoney(user);
-        }
+        public async Task<int> GetAvailableMoney(T user) =>
+            await GetTotalMoney(user) - await GetReservedMoney(user);
 
         public abstract Task<IList<TransactionLog>> PerformTransactions(
```
red031000, line 165: not particularly a fan of this formatting, but I guess it's ok

Felk: 😮

in what way?

red031000: it would probably look better on one line, I don't think it's too long for that?

PR #60, opened 2020-06-25
.NET 5.0

Felk:


red031000:
if this was to be merged, would net5 become required? I assume so due to syntax changes, but do we really want to do this based on an RC and not a stable version?

Felk:
> if this was to be merged, would net5 become required?

yes

> do we really want to do this based on an RC and not a stable version?

It is not a stable version, but it still says this on their homepage:

> Release candidate (RC) releases provide early access to features and are feature complete. These releases are supported for production use since they have a go-live license.

What sdk/runtime do we currently have installed? If it's already 3.1 I'd say let's wait. Otherwise we need to update anyways

PR #61, opened 2020-06-25
add workflow to remove UTF-8 BOMs

Felk:


red031000:
not sure if we need a workflow for this? wouldn't it be better to have a config setting to not insert them (if that even exists)

PR #62, opened 2020-06-26
hook up some static commands

Felk:
some of these commands will need to get cooldowns. I'll add cooldowns in a separate branch and bring them together after one of the two gets merged

### DIFF
--- a/Core/Commands/Definitions/StaticResponseCommands.cs
+++ b/Core/Commands/Definitions/StaticResponseCommands.cs
@@ -29,10 +29,6 @@ namespace Core.Commands.Definitions
                 Description = "Respond with the link to the season rewards webpage."
             },
 
-            new Command("gauntlet", StaticResponse(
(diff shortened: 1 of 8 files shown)


Core/Commands/Definitions/StaticResponseCommands.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using static Core.Commands.CommandUtils;
+
+namespace Core.Commands.Definitions
+{
+    public class StaticResponseCommands : ICommandCollection
+    {
+        // TODO moderators, operators
```
Felk, line 10: I was unsure whether we want to keep those in the config or specify them e.g. in the database. Any thoughts?

Felk:
pls squash if possible, commit history is too noisy for my taste

PR #65, opened 2020-07-08
add common pokemon related models

Felk:


Common/PkmnModels/Status.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    [DataContract]
+    public struct VolatileStatus
+    {
+        [DataMember(Name = "cnf")] public int Cnf { get; set; }
+        [DataMember(Name = "cur")] public bool Cur { get; set; }
+        [DataMember(Name = "foc")] public bool Foc { get; set; }
+        [DataMember(Name = "inf")] public bool Inf { get; set; }
+        [DataMember(Name = "tau")] public bool Tau { get; set; }
+        [DataMember(Name = "tor")] public bool Tor { get; set; }
```
mathfreak231, line 24: I've never seen volatiles except *maybe* confusion be referred to with these 3-letter abbreviations. I can't even guess at what "tau" and "tor" are. Volatile statuses are usually given their full name. This isn't an exhaustive list, either, from what I can tell.

Common/PkmnModels/Stats.cs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    [DataContract]
+    public struct Stats
+    {
+        [DataMember(Name = "hp")] public int Hp { get; set; }
+        [DataMember(Name = "atk")] public int Atk { get; set; }
+        [DataMember(Name = "def")] public int Def { get; set; }
+        [DataMember(Name = "spA")] public int SpA { get; set; }
+        [DataMember(Name = "spD")] public int SpD { get; set; }
```
mathfreak231, line 23: How breaking of a change would it be to make these abbreviations all-lowercase? For reference, Showdown uses lowercase keys in stat tables.

Common/PkmnModels/ElementalType.cs
```diff
@@ -0,0 +1,34 @@
+using System.Runtime.Serialization;
+
+namespace Common.PkmnModels
+{
+    /// Purposely not just named "Type" to avoid confusion with System.Type
+    [DataContract]
+    public enum ElementalType
```
mathfreak231, line 7: How about "PokemonType" instead?

Common/PkmnModels/Pokeset.cs
```diff
@@ -0,0 +1,18 @@
+using System.Collections.Immutable;
+using System.Runtime.Serialization;
+
+namespace Common.PkmnModels
+{
+    [DataContract]
+    public struct Pokeset
+    {
+        public Pokemon Pokemon { get; set; }
```
mathfreak231, line 9: Will this also have compatibility with pokecat's multiple-options functionality? i.e. lists of abilities/items/etc., and a list of lists of moves

Felk: no, that is pokecat's concern. now that I think about it calling the other struct "pokeset" might actually be wrong, since in pokecat's terms a pokeset is exactly that, and not just a pokemon with some additional data. The same thing is probably to remove the Pokeset struct from here, since that will probably stay pokecat's concern anyway.

Common/PkmnModels/Stats.cs
Felk: what does ps call them? just the same but lowercase?

Common/PkmnModels/Status.cs
Felk: I don't know either, I just took whatever pbrengine spat out and put it here 😄
Would making this list exhaustive and giving the things descriptive code names be okay with you? Changing the representations is a bit annoying, since I'd need to hunt down and change all occurance in pbrengine, the old core and the overlay

Common/PkmnModels/Stats.cs
mathfreak231: yes

Common/PkmnModels/Pokeset.cs
mathfreak231: yeah, perhaps

Common/PkmnModels/Status.cs
mathfreak231: Are they even used in the core/overlay currently? I can't think of a place where they would be, except *maybe* in run API, but that's not pbrengine anyways

If we are going to have a use case for these then yes, it's probably worth having any volatile status here. Side effects and field effects, too, why not. Showdown almost always identifies them by their respective move/ability/item name (so "attract" instead of "infatuated")

Common/PkmnModels/Stats.cs
Felk: a quick fulltext search on old core, pbrengine and pokecat shows:
- pokesets would need to be converted, or alternatively stay inconsistent with the ps convention and be converted on instantiation
- code in pokecat needs to be adjusted
- data files need to be adjusted (e.g. pokedex.json)
- documentation needs to be adjusted (examples, specification)
- some injection code in pbrengine needs to be adjusted
- some overlay code needs to be adjusted
- some matchmaker code needs to be adjusted
- API needs to be adjusted (publicly facing breaking change)

Felk: forgot something:
- already persisted data needs to be adjusted, or a change in the data at some point in time needs to be taken into consideration (document it in the api? convert on the fly?)

Common/PkmnModels/Status.cs
Felk: yes, "tau" is used in the calculation of bezier curves, next to "sin" and "pi".

jokes aside, it seems to only use nonvolatile statuses:
```javascript
for (var key in pokeset.status.nonvolatile) {
    let val = pokeset.status.nonvolatile[key];
    if (val) {
        st_url = "/static/status/bg/" + key + ".jpg";
        $sprite.css({"background-image": 'url("' + st_url + '")'});
        nv_count += 1;
    }
}
```

also I found this in the current match data update code:
```python
"status": {
    "nonvolatile": deepcopy(mon["status"]["nonvolatile"]),
    # "volatile": deepcopy(mon["status"]["volatile"]),  # doesn't seem to get updated by pbrengine
},
```

so I think we can safely remove volatile stuff until we actually need it

aaaaaa123456789: If PbrEngine is using confusing three-letter abbreviations, then maybe fix _that_ instead of propagating a bad standard?

Common/PkmnModels/Stats.cs
aaaaaa123456789: I can't really see any benefit of making this change...

Common/PkmnModels/Status.cs
Felk: It isn't, those are currently set by pokecat. to quote ferraro from discord:
> I added those fields to pokecat during pokeset population, see init line 408
> I was also not really sure how to handle those fields though, since they do seem to relate to a particular pokemon instance

The above code was a direct depiction of what went "over the wire", as a starting point and to be able to discuss how things currently are.

Also, sorry if that's not how you wanted to sound, but please don't just comment here to drop superficial passive-aggressive oneliners.

aaaaaa123456789: I get flooded by emails whenever threads like this one get going. Don't blame me for actually checking them out.
That being said, if you know the source of those confusing labels, it's probably best to start there (wherever that source may be) and fix this from the source.

If you think my suggestions are "passive-aggressive one-liners", I'd strongly recommend taking a short break; you're reading stuff that isn't there.

mathfreak231:
my concerns have been resolved I think

PR #66, opened 2020-07-08
fix null issue in mock logger

Felk:


Felk:
I have no idea why this sometimes fails. My best guess is that it relied on some implementation detail of the logging framework. Making it extra nullsafe seems to work

red031000:
very odd why this works, but ok

PR #67, opened 2020-07-10
hook up persistence repositories for executable core

Felk:


Core/Configuration/RootConfig.cs
```diff
@@ -15,6 +15,14 @@ public sealed class RootConfig : ConfigBase
...
         public string? LogPath { get; private set; } = null;
 
+        /* connection details for mongodb */
+        public string MongoDbConnectionUri { get; private set; } = "mongodb://localhost:27017/?replicaSet=rs0";
+        public string MongoDbDatabaseName { get; private set; } = "tpp3";
+
         public IrcConfig Irc { get; private set; } = new IrcConfig();
+
+        /* currency amounts for brand new users (new entries in the database) */
+        public int StartingPokeyen { get; private set; } = 100;
```
red031000, line 25: default is 1000 is it not?

Felk: actually it isn't. 1000 is the tutorial bonus

PR #68, opened 2020-07-10
connect to chat using TwitchLib

Felk:
requires https://github.com/TwitchPlaysPokemon/tpp-core/pull/67 to be merged first.

### DIFF
--- a/Core/Chat/IChat.cs
+++ b/Core/Chat/IChat.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Threading.Tasks;
+using Persistence.Models;
+
+namespace Core.Chat
+{
+    public class MessageEventArgs : EventArgs
+    {
+        public Message Message { get; }
+        public MessageEventArgs(Message message) => Message = message;
+    }
+
+    public interface IChat
+    {
+        event EventHandler<MessageEventArgs> IncomingMessage;
+
+        /// <summary>
+        /// Establishes the connection and keeps it alive until the returned disposable is disposed.
+        /// This method may only ever be called once per instance of this class.
+        /// </summary>
+        /// <returns>A disposable that properly disconnects and cleans up the connection when disposed.</returns>
+        IDisposable EstablishConnection();
+
+        Task SendMessage(string message);
+        Task SendWhisper(User target, string message);
+    }
+}
--- a/Core/Chat/IChat.cs
+++ b/Core/Chat/IChat.cs
@@ -10,7 +10,13 @@ namespace Core.Chat
         public MessageEventArgs(Message message) => Message = message;
     }
 
-    public interface IChat
+    public interface IMessageSender
+    {
+        Task SendMessage(string message);
+        Task SendWhisper(User target, string message);
+    }
+
+    public interface IChat : IMessageSender
     {
         event EventHandler<MessageEventArgs> IncomingMessage;
 
@@ -20,8 +26,5 @@ namespace Core.Chat
         /// </summary>
         /// <returns>A disposable that properly disconnects and cleans up the connection when disposed.</returns>
         IDisposable EstablishConnection();
-
-        Task SendMessage(string message);
-        Task SendWhisper(User target, string message);
     }
 }
--- a/Core/Chat/IChat.cs
+++ b/Core/Chat/IChat.cs
@@ -16,15 +16,13 @@ namespace Core.Chat
         Task SendWhisper(User target, string message);
     }
 
-    public interface IChat : IMessageSender
+    public interface IChat : IMessageSender, IDisposable
     {
         event EventHandler<MessageEventArgs> IncomingMessage;
 
-        /// <summary>
-        /// Establishes the connection and keeps it alive until the returned disposable is disposed.
-        /// This method may only ever be called once per instance of this class.
-        /// </summary>
-        /// <returns>A disposable that properly disconnects and cleans up the connection when disposed.</returns>
-        IDisposable EstablishConnection();
+        /// Establishes the connection.
+        /// All subsequent repeated invocations on this instance will fail.
+        /// The connection gets closed by disposing this instance.
+        void Connect();
     }
 }
--- a/Core/Chat/TwitchChat.cs
+++ b/Core/Chat/TwitchChat.cs
@@ -1,0 +1,177 @@
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Core.Configuration;
+using Microsoft.Extensions.Logging;
+using NodaTime;
+using Persistence.Models;
+using Persistence.Repos;
+using TwitchLib.Client;
+using TwitchLib.Client.Events;
+using TwitchLib.Client.Models;
+using TwitchLib.Communication.Clients;
+using TwitchLib.Communication.Models;
+
+namespace Core.Chat
+{
+    public sealed class TwitchChat : IChat
+    {
+        public event EventHandler<MessageEventArgs> IncomingMessage = null!;
+
+        private readonly ILogger<TwitchChat> _logger;
+        private readonly IClock _clock;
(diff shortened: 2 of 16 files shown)


Core/Configuration/IrcConfig.cs
```diff
@@ -18,6 +18,9 @@ public sealed class IrcConfig : ConfigBase
         public string Username { get; private set; } = "justinfan27365461784";
         public string Password { get; private set; } = "oauth:mysecret";
 
+        // TODO this configurations should probably be in the database instead
+        public IImmutableList<string> OperatorNames { get; private set; } = ImmutableList<string>.Empty;
```
Felk, line 22: Haven't figured out where and how exactly I'd put that, and if putting it in the database even if a desirable thing. E.g. if I get a database dump from stream I wouldn't want that to interfere with my development configuration

Core/Tpp.cs
```diff
@@ -47,7 +58,12 @@ private async Task OnMessage(Message message)
                     .Process(commandName, parts.Skip(1).ToImmutableList(), message);
                 if (result.Response != null)
                 {
-                    _logger.LogDebug($"> {message.User}: {result.Response}");
+                    if (message.MessageSource == MessageSource.Chat)
+                        await _chat.SendMessage($"@{message.User.TwitchDisplayName} {result.Response}");
+                    else if (message.MessageSource == MessageSource.Whisper)
+                        await _chat.SendWhisper(message.User, result.Response);
```
Felk, line 64: for most commands this will work fine. but there will inevitably be commands that want to controle where they post their response. I haven't figured out a way to add this to the command processor that wouldn't needlessly bloat it and instead thought the commands that _do_ need that will just have to depend on the chat and send messages themselves. Does that sound about right?

Core/Chat/IChat.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        /// <summary>
+        /// Establishes the connection and keeps it alive until the returned disposable is disposed.
+        /// This method may only ever be called once per instance of this class.
+        /// </summary>
+        /// <returns>A disposable that properly disconnects and cleans up the connection when disposed.</returns>
+        IDisposable EstablishConnection();
+
+        Task SendMessage(string message);
+        Task SendWhisper(User target, string message);
```
Felk, line 25: might be worth to extract these 2 methods in a base interface `IMessageSender`, in case someone needs to depend on just sending messages and doesn't need to see the other stuff. Commands manually sending messages, as mentioned in another comment, could be one usecase.

Core/Chat/TwitchChat.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            if (_suppressions.Contains(IrcConfig.SuppressionType.Whisper) &&
+                !_suppressionOverrides.Contains(target.SimpleName))
+            {
+                _logger.LogDebug($"(suppressed) >@{target.SimpleName}: {message}");
+                return Task.CompletedTask;
+            }
+            _logger.LogDebug($">@{target.SimpleName}: {message}");
+            _twitchClient.SendWhisper(target.SimpleName, message);
+            return Task.CompletedTask;
```
Felk, line 84: TwitchLib doesn't do async, but the interface felt like it was supposed to be async anyway. That's why it's just returning `Task.CompletedTask` in case you wondered

Core/Chat/TwitchChat.cs
```diff
@@ -0,0 +1,177 @@
...
+        public Task SendMessage(string message)
+        {
+            if (_suppressions.Contains(IrcConfig.SuppressionType.Message) &&
+                !_suppressionOverrides.Contains(_ircChannel))
+            {
+                _logger.LogDebug($"(suppressed) >#{_ircChannel}: {message}");
+                return Task.CompletedTask;
+            }
+            _logger.LogDebug($">#{_ircChannel}: {message}");
+            _twitchClient.SendMessage(_ircChannel, message);
```
Felk, line 70: This will eventually need to be modified to properly support long messages. Probably just cutting it into multiple messages. Right now, twitchlib just drops too long messages with an error log

Core/Commands/Definitions/AdminCommands.cs
```diff
@@ -0,0 +1,40 @@
...
+            _stopToken = stopToken;
+            _operatorNamesLower = operatorNames.Select(s => s.ToLowerInvariant()).ToImmutableHashSet();
+        }
+
+        public IEnumerable<Command> Commands => new[]
+        {
+            new Command("stop", Stop),
+        }.Select(cmd => cmd.WithRestrictedByUser(IsOperator));
+
+        private bool IsOperator(User user) => _operatorNamesLower.Contains(user.SimpleName.ToLowerInvariant());
```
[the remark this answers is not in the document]
red031000: op should be based on user ID, not username due to potential username changes

Core/Tpp.cs
red031000: might be a good idea to address that when/if we come across that, since it might need to be specific to the command how that's implemented

Core/Chat/TwitchChat.cs
red031000: do we want to make these call Task.Run instead?

Core/Commands/CommandExtensions.cs
```diff
@@ -42,5 +44,15 @@ public static Command WithPerUserCooldown(this Command command, Duration duratio
                         : Task.FromResult(new CommandResult()))
             { Aliases = command.Aliases, Description = command.Description };
         }
+
+        public static Command WithRestrictedByUser(this Command command, Func<User, bool> canExecute)
```
red031000, line 48: I'm wondering if it's better to do restrictions based on attributes instead of chaining them, for example
[OperatorOnly] or [ModeratorOnly]

Core/Configuration/IrcConfig.cs
red031000: database would allow the list to change dynamically, but it might also slow stuff down if we wanted to get a list of operators

Core/Commands/CommandExtensions.cs
Felk: I don't know how I would do that without
- complicate the command execution signature, or
- relying on reflection to read those attributes and modify the execution accordingly, which relies on "magic" that the compiler wouldn't see if it suddenly didn't happen anymore

Core/Configuration/IrcConfig.cs
Felk: I wouldn't worry about performance. Might be a good idea to add some sort of generic role system that things like "trusted" can use too. Anyway, can I do this later/in a separate branch?

Core/Commands/Definitions/AdminCommands.cs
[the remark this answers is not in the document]
Felk: if going over persistence, yes. For a temporary solution over a config file fiddling with user ids would be painful

Core/Commands/Definitions/OperatorCommands.cs
```diff
@@ -0,0 +1,58 @@
...
+        public IEnumerable<Command> Commands => new[]
+        {
+            new Command("stop", Stop),
+        }.Select(cmd => cmd.WithCondition(
+            canExecute: ctx => IsOperator(ctx.Message.User),
+            ersatzResult: new CommandResult { Response = "Only operators can use that command" }));
+
+        private bool IsOperator(User user) => _operatorNamesLower.Contains(user.SimpleName);
+
+        private async Task<CommandResult> Stop(CommandContext context)
```
red031000, line 37: when running simultaniously, we should have !stop new/!stop new cancel, or !stop old/!stop old cancel to only stop one of the cores, ofc !stop should stop both if no other args are specified, same with !stop cancel

Core/Configuration/IrcConfig.cs
red031000: yeah that sounds fine to me

PR #70, opened 2020-07-11
fix unstable test TestLogSlowCommand

Felk:
My original fix, adding null checks there, was completely wrong.
~Turns out this issue occurs every time it tried to verify a log call
but there was none. I still don't know why _that_ happens,
but at least adjusting the sleep in the test seems to reliably fix the issue.~
It was a timing issue, and a bug in Moq caused a NullReferenceException
instead of correctly stating that there were no invocations.


Felk:
lol, thanks dependabot

> Regression: NullReferenceException on subsequent setup if expression contains null reference

PR #76, opened 2020-07-11
fix some formatting rules

Felk:
this resolves some situations where dotnet-format did not play well with other tools,
e.g. resharper assuming different defaults,
or unnecessarily replacing all CRLF with LF when git's autocrlf is set to true.
There is still one indentation issue left I cannot fix yet,
see https://github.com/dotnet/format/issues/730 for details on that.


Felk:
Alternatively set `csharp_space_after_cast = false`. I don't mind either so I just set whatever. If someone has a preference we can change it

PR #77, opened 2020-07-12
fix race condition in MongoDB UserRepo.RecordUser

Felk:


Felk:
This hardens the new core for dual-core usage.

I also attempted a different approach: always trying to insert a new user first and always catching the exception. Since users already existing is the 99% case, that would throw an exception almost every time. Quick unscientific measurements showed that recording users then took ~4x more time, so I did it this way to have a clean and exception-free "good path".

PR #82, opened 2020-07-17
make the bot use '/me' and fix message length limits

Felk:


Core/Chat/TwitchChat.cs
```diff
@@ -19,7 +19,10 @@ namespace Core.Chat
     public sealed class TwitchChat : IChat
     {
         public event EventHandler<MessageEventArgs> IncomingMessage = null!;
-        private static readonly MessageSplitter MessageSplitter = new MessageSplitter(maxMessageLength: 500);
+        private static readonly MessageSplitter MessageSplitterRegular = new MessageSplitter(
+            maxMessageLength: 500 - "/me ".Length);
+        private static readonly MessageSplitter MessageSplitterWhisper = new MessageSplitter(
+            maxMessageLength: 500 - "/w ,,,,,''''',,,,,''''',,,,, ".Length);
```
red031000, line 25: again this one really confuses me, why is it how it is?

Felk: the `/w ,,,,,''''',,,,,''''',,,,, ` is a placeholder for some fixed text that gets added to the actual text, so it needs to be subtracted from the total length
I can add a comment clarifying that those ''' and ,,, are just for visuals, for easier counting

red031000: comment would be good

PR #84, opened 2020-07-19
add workflow to publish Core

Felk:


red031000:
I assume that we want this then, if so then all that remains is to set up a system to download (preferably through a batch file) and run the core

Felk:
I thought we decided to just have an SDK installed and run `dotnet run` locally? This was just a proof of concept that it _is_ pretty simple to automate builds if we need to.

red031000:
so if we're not going for this then can I close it?

PR #86, opened 2020-07-21
Pokedex data

Felk:


ArgsParsing/TypeParsers/PkmnSpeciesParser.cs
```diff
@@ -1,38 +1,53 @@
...
                 }
+                if (args.Count >= 2)
+                {
+                    string normalizedNameTwoArgs = NormalizeName(args[0] + ' ' + args[1]);
+                    if (_nameLookup.TryGetValue(normalizedNameTwoArgs, out PkmnSpecies speciesFromTwoArgsName))
+                    {
+                        return Task.FromResult(ArgsParseResult<PkmnSpecies>.Success(
+                            speciesFromTwoArgsName, args.Skip(2).ToImmutableList()));
+                    }
+                }
```
Felk, line 50: this new block has an interesting side-effect: I can put a space in-between a species name that does not actually have a space, e.g. `pid gey`, and it will be found. This feels very odd.
Off the top of my head I can think of 2 possible fixes:
- Use a less generic name normalization. This could be a hardcoded list of some name variations like `mr mime`, `mr-mime`, `mrmime` etc.
- add a special case here: if a pokemon was found while using 2 arguments, check if the found species' name actually contains a space. Otherwise treat it as a failure.

I think the former might be better, as in its current format there are countless other variations that really shouldn't match, e.g. `.:-'P'-:.:-'i'-:.:-'d'-:.:-'g'-:.:-'e'-:.:-'y'-:`

red031000: the former would probably be preferable

PR #100, opened 2020-10-02
ArgsParsing.Tests: use NFluent for exception assertions

Felk:
just an experiment to see how you like fluent assertions

### DIFF
--- a/ArgsParsing.Tests/ArgsParserTest.cs
+++ b/ArgsParsing.Tests/ArgsParserTest.cs
@@ -1,6 +1,7 @@
 using System.Collections.Immutable;
 using ArgsParsing.TypeParsers;
 using ArgsParsing.Types;
+using NFluent;
 using NodaTime;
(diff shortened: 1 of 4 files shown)


red031000:
what's the advantage to this?

Felk:
> what's the advantage to this?

nothing, just readability, and that's subjective

red031000:
do we want to do this then? maybe @m4-used-rollout has an opinion on which one is better

red031000:
can we either decide if we want this or close it

Felk:
I don't know. I like it, I feel like it's up to you liking it as well as m4 probably won't be involved all that much in new-core C# code

red031000:
honestly I'm a bit unsure on it, I like the format, but I don't like that it's another (unnecessary imo) package, and I also don't particularly like the syntax for .And

Felk:
alright, let's not do it. at least for now

PR #107, opened 2020-10-12
generify modes for better extensibility, add dualcore mode

Felk:


Core/Modes/DualcoreMode.cs
```diff
@@ -0,0 +1,12 @@
+using Core.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Core.Modes
+{
+    public class DualcoreMode : ModeBase
```
Felk, line 6: Avoid inheritance, just use an interface and composition

red031000 (approved):

PR #111, opened 2020-10-17
add basic interfaces and structures for match cycles

Felk:


Match/FeatureSet.cs
```diff
@@ -0,0 +1,37 @@
+using System.Collections.Immutable;
+
+namespace Match
+{
+    public enum Generation
+    {
+        Gen1, Gen2, Gen3, Gen4, Gen5, Gen6, Gen7, Gen8
```
red031000, line 7: start enum at 1

Match/FeatureSet.cs
```diff
@@ -0,0 +1,37 @@
+using System.Collections.Immutable;
+
+namespace Match
+{
+    public enum Generation
```
red031000, line 5: put enums in a class so there are no naming conflicts?

Match/IMatchCycle.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// information on what features the match implementation supports
+        public FeatureSet FeatureSet { get; }
+
+        /// Sets up a match for the provided match info,
+        /// and returns a delegate to perform the prepared match.
+        public Task<Perform> SetUp(MatchInfo matchInfo, CancellationToken? token = null);
+
+        /// A delegate that performs a match that was prepared with <see cref="IMatchCycle.SetUp"/>.
+        delegate Task<MatchResult> Perform(CancellationToken? token = null);
```
red031000, line 16: delegate vs event?

Match/MatchInfo.cs
```diff
@@ -0,0 +1,19 @@
...
+
+    public record MatchInfo(
+        IImmutableList<Pokemon> TeamBlue,
+        IImmutableList<Pokemon> TeamRed)
+    {
+        public string? Stage { get; init; }
+    // dotnet-format is drunk, see: https://github.com/dotnet/format/issues/805
+    public string? FieldEffect { get; init; }
+    public float Speed { get; init; } = 1f;
+}
```
red031000, line 18: bracket alignment?

red031000 (changes requested):
match tests?

Match/MatchInfo.cs
Felk: see comment above, dotnet is doing weird stuff. 
If that's preferable I can fix it by marking this file as generated code in the editorconfig file instead, so dotnet-format won't touch it.

Match/FeatureSet.cs
Felk: Is there something also called `Generation` I should be aware of?

Match/IMatchCycle.cs
Felk: not sure what you mean

Felk: The reason this isn't just a second interface method is to make sure nobody could call Perform before SetUp

Match/FeatureSet.cs
red031000: probably should have put this on features instead, incase we want to use it to enable run stuff or smth like that

Match/IMatchCycle.cs
red031000: I mean is an event preferable to a delegate here

Match/MatchInfo.cs
red031000: no that should be fine

Match/IMatchCycle.cs
Felk: after a lengthly discord discussion we agreed that this is confusing and should just be two interface methods

red031000 (approved):

PR #120, opened 2020-10-24
add websocket events for the (legacy) overlay

Felk:


Core.Tests/Overlay/WebsocketBroadcastServerTest.cs
```diff
@@ -0,0 +1,103 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using NUnit.Framework;
+
+namespace Core.Tests.Overlay
+{
+    [Category("IntegrationTest")]
+    public class WebsocketBroadcastServerTest
+    {
+        private const int Port = 53427;
+        private Task _wsServerListen = null!;
```
red031000, line 17: Task?

Core.Tests/Overlay/WebsocketBroadcastServerTest.cs
```diff
@@ -0,0 +1,103 @@
...
+        private WebsocketBroadcastServer _server = null!;
```
red031000, line 18: see above

Core.Tests/Overlay/WebsocketBroadcastServerTest.cs
```diff
@@ -0,0 +1,103 @@
...
+            await _server.Stop();
+            await _wsServerListen;
+        }
+
+        private static async Task<WebsocketMessageStreamClient> CreateConnectedClient()
+        {
+            var url = new Uri($"ws://localhost:{Port}");
+            var wsClient = new WebsocketMessageStreamClient();
+            await wsClient.Connect(url, CancellationToken.None);
+            await Task.Delay(TimeSpan.FromMilliseconds(10)); // wait for the server to accept the connection
```
red031000, line 39: 10ms seems pretty arbitrary, are you sure it'll be enough? I can see it leading to failed tests in the future

Core/Overlay/WebsocketBroadcastServer.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Send a message to all currently connected clients.
+        public async Task Send(string message, CancellationToken cancellationToken)
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(message);
+            await _connectionsSemaphore.WaitAsync(cancellationToken);
+            try
+            {
+                foreach (var connection in _connections)
+                {
+                    await connection.WebSocket.SendAsync(bytes, WebSocketMessageType.Text, endOfMessage: true,
```
red031000, line 60: don't think that this should end if one client has an error, ideally it sends to all that it can, skipping ones with errors (try should be inside the foreach?)

Felk:
> should we be using wss and https instead of ws and http? I know it makes things more complicated but better to be secure?

We won't have any websocket connections that leave localhost. If we had any websockets going over the internet they would be proxied over nginx anyway (internet <-wss-> nginx <--> localhost). I don't think secure connections are required here.

red031000 (approved):

PR #121, opened 2020-10-27
fix crash for people having more than 2^31 money

Felk:


Felk:
this should just _barely_ fit frosted's 2536674963109890646 tokens

red031000 (approved):

PR #133, opened 2020-11-11
Fix #112

### LINKED ISSUE #112: !badges command is not agnostic to the argument order
The !badges command should work with both `!badges <user> <species>` and `!badges <species> <user>`.  
However, `!badges pidgey felkcraft` works, but `!badges felkcraft pidgey` gives the following error:

> No pokemon with the name 'felkcraft' was recognized. Please supply a valid name, or prefix with '#' to supply and pokedex number instead

6100m:
It was actually easier than I expected. I hope to do more PR's of more complex stuff as I gain more skill.
EDIT: Sorry if I referenced 2, that was a accident.

### DIFF
--- a/Core/Commands/Definitions/BadgeCommands.cs
+++ b/Core/Commands/Definitions/BadgeCommands.cs
@@ -47,7 +47,7 @@ namespace Core.Commands.Definitions
 
         public async Task<CommandResult> Badges(CommandContext context)
         {
-            (Optional<User> optionalUse, Optional<PkmnSpecies> optionalSpecies) =
+            (Optional<User> optionalUser, Optional<PkmnSpecies> optionalSpecies) =
                 await context.ParseArgsOptional<User>, <Optional<PkmnSpecies>>();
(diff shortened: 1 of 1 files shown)


aaaaaa123456789:
Doesn't even build.

6100m:
@aaaaaa123456789 Yeah I know, something happened when I swapped it around.
Optional<User>, <ParseArgsOptional<PkmnSpecies>>

Maybe this is the way it wants it?

aaaaaa123456789:
Please test your code locally instead of flooding us with commits that don't even compile, let alone build.

6100m:
I don't have a method to test it actually, yeah apologizes for so many commits.

aaaaaa123456789:
In that case, given that you're clueless about the syntax of the language, I'm going to prevent our inboxes from getting flooded.

6100m:
And yeah, this isn't my code to begin with, I am fixing Issue 112. I am trying one thing after another until it compiles.

PR #142, opened 2020-12-06
forward unprocessed messages via messagequeue

Felk:
this is the first step to allow for easier dualcore cooperation:
- new core takes on the responsibility of generic message logging
- new core writes its own command log entries instead of relying on the old core's stubs to do that
- new core writes messages it did not already process itself to a database messagequeue, where old core can pick it up. https://github.com/TwitchPlaysPokemon/tpp/pull/468 needs to be merged first for this to work

this is the "partner" branch of https://github.com/TwitchPlaysPokemon/tpp/pull/470

TODO:
- [x] add some tests
- [x] figure out how to allow the old core to ditch its command stubs, as right now they are still required for the !help command, and maybe for other stuff I'm unaware of?
- [x] give it a proper full-fledged dual-core testrun
- [x] merge https://github.com/TwitchPlaysPokemon/tpp/pull/468
- [x] merge https://github.com/TwitchPlaysPokemon/tpp/pull/470


Felk:
regarding the `!help` command, it's probably easiest to just move that to the new core and have all commands' descriptions that only exist in new core be hardcoded in the command for now

Felk:
> figure out how to allow the old core to ditch its command stubs, as right now they are still required for the !help command, and maybe for other stuff I'm unaware of?

I don't think there's anything besides that, so I just went ahead and hardcoded all of the old core's command's descriptions into the help command here

Felk:
> give it a proper full-fledged dual-core testrun

locally, everything worked fine. I'm gonna check this one as well.

Core/Chat/TwitchChat.cs
```diff
@@ -179,7 +179,10 @@ private async void WhisperReceived(object? sender, OnWhisperReceivedArgs e)
                 fromMessage: true,
                 updatedAt: _clock.GetCurrentInstant()
             ));
-            var message = new Message(user, messageText, source);
+            // TODO whispers don't have "RawIrcMessage" yet.
+            // The required change is merged, but not released yet: https://github.com/TwitchLib/TwitchLib.Client/pull/159
+            var rawIrcMessage = (twitchLibMessage as ChatMessage)?.RawIrcMessage ?? string.Empty;
```
red031000, line 184: does this mean that tags won't be available for this? that seems like it'd potentially cause issues

Persistence/Models/CommandLog.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public sealed record CommandLog(
+        string Id,
+        string UserId,
+        string Command,
+        IReadOnlyList<string> Args,
+        Instant Timestamp,
+        string? Response)
+    {
+        public bool Equals(CommandLog? other)
```
red031000, line 16: also override operator ==?

red031000 (changes requested):
few small things, lack of whisper tags can cause problems with DB logging

Core/Chat/TwitchChat.cs
Felk: actually this means that whispers cannot be forwarded at all... I totally forgot about that. This probably just requires a TwitchLib release

Felk: they did have a release, I just included the wrong dependency. Fixed it

red031000 (approved):

PR #144, opened 2020-12-11
add basic matchloop

Felk:


Core/Modes/Matchmode.cs
```diff
@@ -1,38 +1,118 @@
...
+            object winnerForOverlay = result.Winner switch { Side.Blue => 0, Side.Red => 1, _ => "draw" };
+            await _overlayConnection.Send(new MatchOverEvent { MatchResult = winnerForOverlay });
+
+            await Task.Delay(_matchmodeConfig.ResultDuration.ToTimeSpan());
+            await _overlayConnection.Send(new ResultsFinishedEvent());
+        }
+
         public void Cancel()
         {
             // once the mainloop is not just busylooping, this needs to be replaced with something
```
Felk, line 118: this needs to be done now

red031000: so do you want to do this before I review or have a different PR for it?

Felk: you can already look at it, but nothings blocking me from fixing that, I'm just procrastinating.
I'll likely just pass a cancellation token to Loop()

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+                {
+                    await Loop(_loopCancelToken.Token);
+                }
+                catch (OperationCanceledException)
+                {
+                    _logger.LogInformation("Match loop cancelled");
+                }
             }
+            await _broadcastServer.Stop();
+            await overlayWebsocketTask;
```
red031000, line 58: I don't really understand why you listen after you've stopped it?

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
             _logger.LogInformation("Matchmode ended");
         }
 
+        private async Task Loop(CancellationToken cancellationToken)
```
red031000, line 62: I assume that this is a testing version and therefore is not intended for deployment

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+            await _overlayConnection.Send(new MatchCreatedEvent(), cancellationToken);
+            await _overlayConnection.Send(new MatchBettingEvent(), cancellationToken);
+            await _overlayConnection.Send(new MatchModesChosenEvent(), cancellationToken); // TODO
+            await _overlayConnection.Send(new MatchSettingUpEvent
+            {
+                MatchId = 1234,
+                Teams = teams,
+                BettingDuration = _matchmodeConfig.DefaultBettingDuration.TotalSeconds,
+                RevealDuration = 0,
+                Gimmick = "speed",
```
red031000, line 82: idk if gimicks, switching and battle style as strings is a good idea, I'd rather them be enums

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+                Switching = "never",
+                BattleStyle = "singles",
+                InputOptions = new InputOptions
+                {
+                    Moves = new MovesInputOptions
+                    {
+                        Policy = "always",
```
red031000, line 89: again, enum

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+                        Permitted = ImmutableList.Create("a", "b", "c", "d")
+                    },
+                    Switches = new SwitchesInputOptions
+                    {
+                        Policy = "never",
```
red031000, line 94: see above

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+                        Permitted = ImmutableList<string>.Empty,
+                        RandomChance = 0
+                    },
+                    Targets = new TargetsInputOptions
+                    {
+                        Policy = "disabled",
```
red031000, line 100: see above

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+                        Permitted = ImmutableList<string>.Empty,
+                        AllyHitChance = 0
+                    },
+                },
+                BetBonus = 35,
+                BetBonusType = "bet",
```
red031000, line 106: enum

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+            }, cancellationToken);
+
+            Duration bettingBeforeWarning = _matchmodeConfig.DefaultBettingDuration - _matchmodeConfig.WarningDuration;
+            await Task.Delay(bettingBeforeWarning.ToTimeSpan(), cancellationToken);
+            await _overlayConnection.Send(new MatchWarningEvent(), cancellationToken);
+
+            await Task.Delay(_matchmodeConfig.WarningDuration.ToTimeSpan(), cancellationToken);
+            await setupTask;
```
red031000, line 114: should be before here, way before

red031000: i.e. during before warning

Core/Modes/Matchmode.cs
```diff
@@ -1,43 +1,132 @@
...
+            Task<MatchResult> performTask = match.Perform(cancellationToken);
+            await _overlayConnection.Send(new MatchPerformingEvent { Teams = teams }, cancellationToken);
+
+            MatchResult result = await performTask;
+            object winnerForOverlay = result.Winner switch { Side.Blue => 0, Side.Red => 1, _ => "draw" };
```
red031000, line 119: not too sure about object here, but again idk if anything else would work

red031000 (changes requested):

Core/Modes/Matchmode.cs
Felk: I'm not listening, I'm awaiting it to cleanly shut down

Core/Modes/Matchmode.cs
Felk: the whole matchmode is not ready for deployment any time soon. this method in particular isn't a test method, rather it's contents are too hardcoded still

Core/Modes/Matchmode.cs
Felk: I don't think so. The setup is meant to be able to run concurrently to the entire betting time. It is awaited as late as possible, which is just before the match is supposed to begin.

Core/Modes/Matchmode.cs
Felk: Is there even a well-defined list of gimmicks? I think it's just a huge configuration file I'm not familiar with

Core/Modes/Matchmode.cs
Felk: I suggest ignoring the bet bonus stuff for now, because I'm not sure I want to keep bet bonuses the way they currently are anyway. I think it's better if the two kinds of bet bonuses were distinct fields like `bet_bonus` and `profit_bonus` instead of one field whose meaning gets interpreted based on the contents of another field. And that other field doesn't even get persisted right now :(

red031000 (approved):

PR #154, opened 2021-01-09
add simple runmode

Felk:


Felk:
I should also add instructions to the readme

m4-used-rollout (approved):

PR #170, opened 2021-02-03
add emoteonly[on] and emoteonlyoff commands

red031000:
note: with this PR moderator names will also need to be added to the core config, OPs and mods need to be moved to the db eventually however


Common/ListExtensions.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Common
+{
+    public static class ListExtensions
+    {
+        public static List<T> AddRangeReturn<T>(this List<T> value, IEnumerable<T> collection)
+        {
+            value.AddRange(collection);
+            return value;
+        }
```
Felk, line 11: that's not worth an extension method imo, especially since you can "flatten" lists of lists with SelectMany

Core/Commands/Definitions/ModeratorCommands.cs
```diff
@@ -0,0 +1,52 @@
...
+using Common;
+using Core.Chat;
+using Persistence.Models;
+
+namespace Core.Commands.Definitions
+{
+    public class ModeratorCommands : ICommandCollection
+    {
+        private readonly ImmutableHashSet<string> _moderatorNamesLower;
+        private IChatModeChanger _changer;
```
Felk, line 14: readonly

Felk (changes requested):

Core/Chat/IChatModeChanger.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading.Tasks;
+
+namespace Core.Chat
+{
+    public interface IChatModeChanger
```
Felk, line 5: If this is only ever going to be used for the commands you could put it in the command collection's file to keep it scoped more tightly and have fewer files

Felk: Actually maybe not do that, that'd make TwitchChat directly depend on the command

Felk (approved):

PR #171, opened 2021-02-04
prefix all projects with "TPP"

Felk:
about the amount of changes I expected. Unfortunately if this is done it's gonna be annoying for everyone locally, since all untracked files (configs, and on stream pc 1 batch file) will suddenly be at the wrong place


Felk:
Also conflicts for every other branch. But if we want to do this, better now than later

PR #173, opened 2021-02-06
dualcore messagequeue: revert to only sending PRIVMSG and WHISPER

Felk:


Felk:
see also https://github.com/TwitchPlaysPokemon/tpp/pull/478

red031000 (approved):

PR #176, opened 2021-02-06
ArgsParsing: replace int parser with sign specific variants

Felk:
In preparation for some future chat commands that need to
differentiate between signed, unsigned, and positive numbers.
This should prevent some bugs where a command author wasn't
explicit about a parameter's bounds and forgot to check it.

### DIFF
--- a/TPP.ArgsParsing.Tests/ArgsParserTest.cs
+++ b/TPP.ArgsParsing.Tests/ArgsParserTest.cs
@@ -12,10 +12,10 @@ namespace TPP.ArgsParsing.Tests
         public void TestTooFewArguments()
         {
             var argsParser = new ArgsParser();
-            argsParser.AddArgumentParser(new NonnegativeIntParser());
+            argsParser.AddArgumentParser(new NonNegativeIntParser());
 
             var ex = Assert.ThrowsAsync<ArgsParseFailure>(() => argsParser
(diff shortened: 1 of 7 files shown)


TPP.ArgsParsing.Tests/ArgsParserTest.cs
```diff
@@ -12,21 +12,21 @@ public class ArgsParserTest
         public void TestTooFewArguments()
         {
             var argsParser = new ArgsParser();
-            argsParser.AddArgumentParser(new IntParser());
+            argsParser.AddArgumentParser(new NonnegativeIntParser());
```
red031000, line 15: Unsigned int is a lot better imo, or at least go for absolute int

red031000 (changes requested):

TPP.ArgsParsing.Tests/ArgsParserTest.cs
Felk: I wouldn't use "unsigned int" because that's misleading. It's still a regular signed int, the nonnegative refers to its validation, not its data type. "absolute int" sounds weird and non-descriptive to me, but maybe I just haven't heard of "absolute" being used in this context yet

red031000: https://en.wikipedia.org/wiki/Absolute_value

red031000 (approved):

PR #183, opened 2021-02-26
add console chat, allow for multiple chat connections

Felk:
This requires local config files to be updated accordingly


TPP.Core/README.md
```diff
@@ -15,17 +15,30 @@ It will tell you what file you need and how to create it.
 ```
 dotnet run -- testconfig
 ```
-You need to customize a few configurations:
-- `Chat.Username` and `Chat.Password` contain the credentials of some Twitch account that will be the chat bot.
-   You can obtain an oauth token from [here](https://twitchapps.com/tmi/)
+For local testing, you need to remove all but the `console` entry from the `Chat.Connections` config.
+You will be able to use the console as a simulated chat.
```
Felk, line 19: I'd love for it to run with a completely untouched configuration, but the json schema generation doesn't quite manage to handle the polymorphic chat connection classes. So I decided to deliberately put 1 of each possible connection (currently "console" and "twitch") in the default config, so the default config can just be used as a template.

Felk:
I'll merge this by Friday unless someone chimes in

PR #190, opened 2021-03-02
ArgsParsing AnyOrderParser: fix handling of arguments

Felk:
the parser had two issues:
- it assumed every type being parsed would consume exactly one argument,
  which is not always true, e.g. Optional can have 0 and Instant can have 2
- it dropped any possibly spare arguments that come after the AnyOrder argument

This was fixed by permuting the types instead of the arguments, and subsequently
just passing the args as-is to the ParseRaw call.


Felk:
I will merge this on Saturday because https://github.com/TwitchPlaysPokemon/tpp-core/pull/191 is blocked by this

PR #191, opened 2021-03-02
add `giftbadge` and `transferbadge` commands

Felk:
`giftbadge` is ported from old core, while `transferbadge` is new and allows operators to gift badges remotely.

This branch is based on https://github.com/TwitchPlaysPokemon/tpp-core/pull/190, so that should be merged first.


Felk:
I will merge this on Saturday because this feature is being awaited.

PR #197, opened 2021-03-08
Pokedex extensions

Fenris2142:
I abandoned the Enum idea, as "-" were not supported, so I used string-constants instead to define the mode names.

### DIFF
--- a/TPP.Core/Commands/Definitions/BadgeCommands.cs
+++ b/TPP.Core/Commands/Definitions/BadgeCommands.cs
@@ -56,20 +56,20 @@ namespace TPP.Core.Commands.Definitions
         private readonly IUserRepo _userRepo;
         private readonly IMessageSender _messageSender;
         private readonly HashSet<PkmnSpecies>? _whitelist;
-        private readonly PokedexData _pokedexData;
+        private readonly IImmutableSet<TPP.Common.PkmnSpecies> _KnownSpecies;
 
         public BadgeCommands(
             IBadgeRepo badgeRepo,
             IUserRepo userRepo,
             IMessageSender messageSender,
-            PokedexData pokedexData,
+            IImmutableSet<TPP.Common.PkmnSpecies> KnownSpecies,
             HashSet<PkmnSpecies>? whitelist = null
         )
         {
             _badgeRepo = badgeRepo;
             _userRepo = userRepo;
             _messageSender = messageSender;
-            _pokedexData = pokedexData;
+            _KnownSpecies = KnownSpecies;
             _whitelist = whitelist;
         }
 
@@ -167,7 +167,7 @@ namespace TPP.Core.Commands.Definitions
 
             if (mode.Equals(PokedexModeMissing))
             {
-                IEnumerable<PkmnSpecies> missingList = _pokedexData.KnownSpecies.Except(numBadgesPerSpecies.Keys);
+                IEnumerable<PkmnSpecies> missingList = _KnownSpecies.Except(numBadgesPerSpecies.Keys);
                 // Seems like PokedexData is not sorted. So we have to sort the list.
                 IEnumerable<string> badgesFormatted = missingList.OrderBy(entry => entry.Id).Select(entry => $"{entry}");
                 return new CommandResult
--- a/TPP.Core/Commands/Definitions/BadgeCommands.cs
+++ b/TPP.Core/Commands/Definitions/BadgeCommands.cs
@@ -168,8 +168,7 @@ namespace TPP.Core.Commands.Definitions
             if (mode.Equals(PokedexModeMissing))
             {
                 IEnumerable<PkmnSpecies> missingList = _knownSpecies.Except(numBadgesPerSpecies.Keys);
-                // Seems like PokedexData is not sorted. So we have to sort the list.
-                IEnumerable<string> badgesFormatted = missingList.OrderBy(entry => entry.Id).Select(entry => $"{entry}");
+                IEnumerable<string> badgesFormatted = missingList.Select(entry => $"{entry}");
                 return new CommandResult
                 {
                     Response = isSelf
--- a/TPP.Core/Commands/Definitions/BadgeCommands.cs
+++ b/TPP.Core/Commands/Definitions/BadgeCommands.cs
@@ -13,11 +13,11 @@ namespace TPP.Core.Commands.Definitions
     public class BadgeCommands : ICommandCollection
     {
         // using an enum didn't work out, because no "-" characters allowed in enums. So go with const strings instead, or change modes to camel-case
-        private const string PokedexModeComplementFrom      = "complement-from";
+        private const string PokedexModeComplementFrom = "complement-from";
         private const string PokedexModeComplementFromDupes = "complement-from-dupes";
-        private const string PokedexModeMissing             = "missing";
-        private const string PokedexModeDupes               = "dupes";
-        private const string PokedexModeModes               = "modes";
+        private const string PokedexModeMissing = "missing";
+        private const string PokedexModeDupes = "dupes";
+        private const string PokedexModeModes = "modes";
 
         public IEnumerable<Command> Commands => new[]
         {
@@ -56,16 +56,20 @@ namespace TPP.Core.Commands.Definitions
         private readonly IUserRepo _userRepo;
         private readonly IMessageSender _messageSender;
         private readonly HashSet<PkmnSpecies>? _whitelist;
+        private readonly PokedexData _pokedexData;
 
         public BadgeCommands(
             IBadgeRepo badgeRepo,
             IUserRepo userRepo,
             IMessageSender messageSender,
-            HashSet<PkmnSpecies>? whitelist = null)
+            PokedexData pokedexData,
+            HashSet<PkmnSpecies>? whitelist = null
+        )
         {
             _badgeRepo = badgeRepo;
             _userRepo = userRepo;
             _messageSender = messageSender;
+            _pokedexData = pokedexData;
             _whitelist = whitelist;
         }
 
@@ -141,14 +145,12 @@ namespace TPP.Core.Commands.Definitions
 
         public async Task<CommandResult> Pokedex(CommandContext context)
         {
-            (Optional<User> optionalUser, Optional<string> mode_, Optional<User> optionalUser2) =
-                await context.ParseArgs<Optional<User>, Optional<string>,Optional<User>>();
+            (Optional<User> optionalUser, Optional<string> optionalMode, Optional<User> optionalCompareUser) =
+                await context.ParseArgs<Optional<User>, Optional<string>, Optional<User>>();
             User user = optionalUser.IsPresent ? optionalUser.Value : context.Message.User;
             bool isSelf = user == context.Message.User;
 
-            // Have to additionally check if the string is not empty, otherwise "!dex " would jump right into the else - branch.
-            //   Not sure if this is an actual issue when connected to twitch
-            if( !mode_.IsPresent || string.IsNullOrWhiteSpace( mode_.ToString() ) )
+            if (!optionalMode.IsPresent)
             {
                 int numUniqueSpecies = (await _badgeRepo.CountByUserPerSpecies(user.Id)).Count;
                 return new CommandResult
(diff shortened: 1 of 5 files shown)


TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
-            User user = (await context.ParseArgs<Optional<User>>()).OrElse(context.Message.User);
+            (Optional<User> optionalUser, Optional<string> mode_, Optional<User> optionalUser2) =
+                await context.ParseArgs<Optional<User>, Optional<string>,Optional<User>>();
+            User user = optionalUser.IsPresent ? optionalUser.Value : context.Message.User;
             bool isSelf = user == context.Message.User;
-            int numUniqueSpecies = (await _badgeRepo.CountByUserPerSpecies(user.Id)).Count;
-            return new CommandResult
+
+            // Have to additionally check if the string is not empty, otherwise "!dex " would jump right into the else - branch.
+            //   Not sure if this is an actual issue when connected to twitch
```
Felk, line 150: This sounds like an issue I could fix in the argument parsing code by collapsing and trimming whitespace. I'll try that

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
-            };
+                int numUniqueSpecies = (await _badgeRepo.CountByUserPerSpecies(user.Id)).Count;
+                return new CommandResult
+                {
+                    Response = isSelf
+                        ? $"You have collected {numUniqueSpecies} distinct Pokémon badge(s)"
+                        : $"{user.Name} has collected {numUniqueSpecies} distinct Pokémon badge(s)"
+                };
+            }
+            else
```
Felk, line 161: Since the if-case returns, you can omit the else-block and avoid the additional layer of indentation

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+            {
+                string mode = mode_.ToString();
```
Felk, line 163: prefer `.Value` over the Optional's `ToString` method

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                ImmutableSortedDictionary<PkmnSpecies, int> numBadgesPerSpecies =
+                    await _badgeRepo.CountByUserPerSpecies( user.Id );
+
+                if( mode.Equals( PokedexModeMissing ))
+                {
+                    IEnumerable<PkmnSpecies> missingList = PokedexData.Load().KnownSpecies.Except( numBadgesPerSpecies.Keys );
```
Felk, line 169: `PokedexData.Load()` is an expensive call. please pass the already-loaded pokedex data into this class' constructor instead. See https://discord.com/channels/148079346685313034/436252785873387540/817502346145562676 and I can also assist you with that if needed

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                            Response = $"Mode {PokedexModeComplementFromDupes} requires a second user argument"
+                        };
+                    }
+                    else
+                    {
+                        var dupeList = numBadgesPerSpecies.Where( kvp => kvp.Value > 1 );
+                        ImmutableSortedDictionary<PkmnSpecies, int> myNumBadgesPerSpecies =
+                            await _badgeRepo.CountByUserPerSpecies(optionalUser2.Value.Id);
+
+                        var intersectList = dupeList.Except( myNumBadgesPerSpecies );
```
Felk, line 219: looks like you implemented this opposite to how I expected it. That's fine, but the name "complement-from" is probably wrong then. I expected "A complement-from B" to print all badges B can provide for A, but it calculates the badges A can provide for B. Possibly just use "complements" instead of "complement-from" or something

Felk: the term intersection is technically wrong here, since it's a difference. Think venn-diagram: the intersection is the thing in the middle, but the difference is one of the sides.

Felk: imo using "my" in these variables like `myNumBadgesPerSpecies` is confusing since it doesn't necessarily reference the user issuing the command but any user. Something like `compareUser` might fit better

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                            {
+                                Response = $"{user.Name} does not own any duplicate Pokémon badges {optionalUser2.Value.Name} is missing"
+                            };
+                        }
+                        else
+                        {
+                            IEnumerable<string> badgesFormatted = intersectList.Select( kvp => $"{kvp.Key}" );
+                            return new CommandResult
+                            {
+                                Response = $"{user.Name} is owning missing duplicates of the following badge(s): {string.Join(", ", badgesFormatted)}",
```
Felk, line 232: message is missing whose user the dupes are for

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                        }
+                        else
+                        {
+                            IEnumerable<string> badgesFormatted = intersectList.Select( kvp => $"{kvp.Key}" );
+                            return new CommandResult
+                            {
+                                Response = $"{user.Name} is owning the following missing badge(s): {string.Join(", ", badgesFormatted)}",
+                                ResponseTarget = ResponseTarget.WhisperIfLong
+                            };
+                        }
```
Felk, line 268: some of the comments apply to multiple modes

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
 
         public async Task<CommandResult> Pokedex(CommandContext context)
         {
-            User user = (await context.ParseArgs<Optional<User>>()).OrElse(context.Message.User);
+            (Optional<User> optionalUser, Optional<string> mode_, Optional<User> optionalUser2) =
```
Felk, line 144: name suggestion: `optionalMode` to be consistent with the other arguments and avoid the `_`

Felk: name suggestion: `optionalCompareUser` to give it more meaning than just appending `2`

Felk (changes requested):
In addition to these changes, once you make another change the format pipeline will run and fail. It will tell you to install and run dotnet-format. That's to ensure the codebase has consistent formatting. If dotnet-format does something you disagree with, we may be able to reconfigure it in the `.editorconfig` file. But most stuff, like spaces inside `if`s, will need to be changed just to be consistent.

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                        Response = isSelf
+                            ? $"You are currently missing the following badge(s): {string.Join(", ", badgesFormatted)}"
+                            : $"{user.Name} is currently missing the following badge(s): {string.Join(", ", badgesFormatted)}",
+                        ResponseTarget = ResponseTarget.WhisperIfLong
+                    };
+                }
+                else if( mode.Equals( PokedexModeDupes ) )
+                {
+                    var dupeList = numBadgesPerSpecies.Where(kvp => kvp.Value > 1);
+                    if( !dupeList.Any() )
```
Felk, line 183: This is probably a bug. `dupeList` is an `IEnumerable<string>`. `IEnumerable` can only be enumerated once. `Any()` starts enumerating it, possibly causing the first element to be gone for the `dupeList.Select(...)` later on.

Easy solution for these cases is to turn it into a list first using `.ToList()`.

The same applies to basically any calls to `.Any()` in your changes

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,15 +141,150 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                        }
+                    }
+                }
+                else if( mode.Equals( PokedexModeModes ) )
+                {
+                    return new CommandResult
+                    {
+                        Response =  $"Supported modes are \"{PokedexModeDupes}\": Show duplicate badges, \"{PokedexModeMissing}\": Show missing badges, "
+                        + "\"" + PokedexModeComplementFrom + "\": Compares missing badges from User A with owned badges from User B, "
+                        + "\"" + PokedexModeComplementFromDupes + "\": Compares missing badges from User A with owned duplicate badges from User B"
```
Felk, line 277: - you could avoid the escaped quotes by using single quotes inside instead
- you can use string formatting for lines 2 and 3 too. looks weird that you did use it in the first but not the other lines.

but this entire this might be easier to look at using verbatim strings (`@`-prefixed) instead. you might want to check those out.

Felk (changes requested):

Felk:
```
!dex felkbot complement-from felkcraft
@admin Felkbot is owning the following missing badge(s): #001 Bulbasaur, #002 Ivysaur, #016 Pidgey
!badges felkcraft
@admin The reply has been whispered to you.
>User(console-admin/admin): FelkCraft's badges: 1x #004 Charmander, 1x #005 Charmeleon, 1x #007 Squirtle, 1x #010 Caterpie, 1x #011 Metapod, 1x #013 Weedle, 1x #014 Kakuna, 2x #015 Beedrill, 110629x #016 Pidgey, ...
```
I'm not sure why it thinks I'm missing Pidgey

TPP.Core/Commands/Definitions/BadgeCommands.cs
Felk: yep, wasn't too hard. See https://github.com/TwitchPlaysPokemon/tpp-core/pull/199

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: dyschromatopsia and reading too much tutorials where people use variables starting with "my". I hate those too

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: I'll rework that code. Output both users first and then list the result

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: Wrong usage of Singleton. Thats embarrasing

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: Removed

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: Makes sense. I removed every **else** since all of them were hiding behind return - statements

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: lol, forget the last comment. That was just me being pepega. I passed the pokedexData - MemberVariable to BadgeCommands as you suggested.

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: I fixed the any() stuff by converting everything to lists. I will be more careful with "Enumerable" Lists in the future

Fenris2142:
Regarding your Pidgey situation.. I found a bug in my code. I compared 2

**ImmutableSortedDictionary<PkmnSpecies, int>**

with the **except** - Function. This function returned a result if both players Own the Pokemon (key) but with different number of owned badges (value).

I didn't found the issue during testing, because my testdata wasn't good enough honestly.

Luckily fixing it was rather easy:

`                var intersectList = dupeList.Keys.Except(compareUser.Keys).ToList();
`
Just use the "Keys" instead of comparing the entire map

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: I can NOT get myself to use verbatim strings. the code looks awful after that. But I got the escape quotes to work after a tip from you which is fine

Felk: sure

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: switching A and B was quite some trial and error, but it should work as intended now.

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,14 +145,127 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
 
         public async Task<CommandResult> Pokedex(CommandContext context)
         {
-            User user = (await context.ParseArgs<Optional<User>>()).OrElse(context.Message.User);
+            (Optional<User> optionalUser, Optional<string> optionalMode, Optional<User> optionalCompareUser) =
+                await context.ParseArgs<Optional<User>, Optional<string>, Optional<User>>();
+            User user = optionalUser.IsPresent ? optionalUser.Value : context.Message.User;
```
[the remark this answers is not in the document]
Fenris2142: changing it to .OrElse, because I didn't knew that syntax and it's good to have some varierty to look at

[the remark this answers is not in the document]
Felk: `Optional` is a custom type inspired by Java's `Optional`, so not a built-in. Still good to know that exists I suppose

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,14 +145,127 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+                    return new CommandResult
+                    {
+                        Response = $"Mode {PokedexModeComplementFromDupes} requires a second user argument"
+                    };
+                }
+                ImmutableSortedDictionary<PkmnSpecies, int> compareUser =
+                    await _badgeRepo.CountByUserPerSpecies(optionalCompareUser.Value.Id);
+                var compareUserDupeList = compareUser.Where(kvp => kvp.Value > 1).ToImmutableSortedDictionary();
+
+                var differenceList = compareUserDupeList.Keys.Except(numBadgesPerSpecies.Keys).ToList();
```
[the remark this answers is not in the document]
Fenris2142: Looking good and should still be sorted to my understanding

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -134,14 +145,127 @@ public async Task<CommandResult> SelectBadge(CommandContext context)
...
+            }
+
+            string mode = optionalMode.Value;
+            ImmutableSortedDictionary<PkmnSpecies, int> numBadgesPerSpecies =
+                await _badgeRepo.CountByUserPerSpecies(user.Id);
+
+            if (mode.Equals(PokedexModeMissing))
+            {
+                IEnumerable<PkmnSpecies> missingList = _KnownSpecies.Except(numBadgesPerSpecies.Keys);
+                // Seems like PokedexData is not sorted. So we have to sort the list.
```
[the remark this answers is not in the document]
Fenris2142: So you want to change this part? `
        private PokedexData()
        {
            KnownSpecies = LoadSpeciesNames().ToImmutableHashSet();
        }`

Unfortunately I'm unsure about which Collection types should be used in which situations. I just wrote the comment there, but I guess it makes sense that a HashSet is not sorted in any way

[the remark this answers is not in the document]
Felk: can I just paste patches as comments and github can apply them...?
```patch
Index: TPP.Core/PokedexData.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/TPP.Core/PokedexData.cs b/TPP.Core/PokedexData.cs
--- a/TPP.Core/PokedexData.cs	(revision 24f7494677330df2e1e20dddf920388786848582)
+++ b/TPP.Core/PokedexData.cs	(date 1615568330788)
@@ -11,7 +11,7 @@
 {
     public class PokedexData
     {
-        public IImmutableSet<PkmnSpecies> KnownSpecies { get; }
+        public ImmutableSortedSet<PkmnSpecies> KnownSpecies { get; }
 
         private static readonly (string, string)[] NameNormalizations =
         {
@@ -58,7 +58,7 @@
 
         private PokedexData()
         {
-            KnownSpecies = LoadSpeciesNames().ToImmutableHashSet();
+            KnownSpecies = LoadSpeciesNames().ToImmutableSortedSet();
         }
 
         /// Construct an instance with all the pokedex data loaded,
```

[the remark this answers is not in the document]
Felk: aw only syntax highlighting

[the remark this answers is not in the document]
Fenris2142: I have checked all usages/occurances of PokedexData and PokedexData.KnownSpecies (most of them were in testing modules tbf) 
Did not find anything that would speak against changing the datatype. Also I removed the "OrderBy" at "!dex missing" because it actually caused Fakemons to get printed all over the place

Felk (approved):

PR #202, opened 2021-03-12
"is owning" sounds a bit akward , doesn't it?

FrostedGeulleisia:


aaaaaa123456789:
The good old rule still holds: whenever you correct someone's writing, you'll make at least one more mistake than the number of mistakes you're correcting :P

(In your case, you misspelled "awkward" and added a space before a comma.)

FrostedGeulleisia:
I typoed it the first time, then tried to correct it but typoed it the same way again and that's why the space is there.

PR #205, opened 2021-03-16
ArgsParsing: add `ManyOf<T>` parser for parsing lists, use it to allow spaces in 'reason' arguments for operator commands

Felk:


TPP.ArgsParsing/TypeParsers/ManyOfParser.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <br/>
+    /// The parser has a few restrictions:
+    /// <ul>
+    ///   <li>
+    ///   It is assumed that each list entry consumes at least one argument, so it will not work with e.g. Optional.
+    ///   </li>
+    ///   <li>
+    ///   It just greedily parses as many arguments as possible. When used in a way where parsing may be ambiguous to
+    ///   any subsequent types, the parser will consume arguments that were meant for that subsequent type.
+    ///   Since there is no backtracking for successes, overall parsing will then fail because parsing the subsequent
```
red031000, line 24: might be worth implementing at the cost of slowing down parsing of commands that use ManyOf

red031000 (approved):

Issue #79, opened 2020-07-15
user parser should also accept display name


red031000:
useful for foreign display names

Felk:
I'll leave this issue open for grabs for someone who hasn't done anything in this codebase yet, as it's a good issue to get started with

Felk:
fixed by a46fdc63b2b908f1962319a33b076e3e9d6b76e8

PR #207, opened 2021-03-20
User parser accepts (foreign) display names #79

Dnantz:


Felk:
Thank you for this contribution! Since this was meant as a first impression, I let the other guys know accordingly

PR #210, opened 2021-03-20
add subscription handling

Felk:
This is almost a direct port of the old core's behaviour. A few things are still blocking a merge:

- [X] Check linked accounts to not give out gift sub reward tokens for those (requires https://github.com/TwitchPlaysPokemon/tpp-core/pull/212)
- [X] Send overlay events (requires https://github.com/TwitchPlaysPokemon/tpp-core/pull/211)
- [X] ~Hook up emote service to parse emotes and include them in the overlay event~ just using the emotes twitch sends along for now
- [X] Better take another close look at the old core to see if I missed anything else
- [x] Disable sub handling in old core (https://github.com/TwitchPlaysPokemon/tpp/pull/487)
- [x] secondary websocket connection in old core (https://github.com/TwitchPlaysPokemon/tpp/pull/483)

PR #213, opened 2021-03-22
Balancereset

Dnantz:
Port balancereset from old core.

Whispers are not yet sent to users when this happens, however this requires some changes on how chats handled.

### DIFF
--- a/TPP.Core/Modes/Matchmode.cs
+++ b/TPP.Core/Modes/Matchmode.cs
@@ -9,6 +9,8 @@ using TPP.Core.Configuration;
 using TPP.Core.Overlay;
 using TPP.Core.Overlay.Events;
 using TPP.Match;
+using TPP.Persistence.Models;
+using TPP.Persistence.Repos;
 
 namespace TPP.Core.Modes
 {
@@ -22,6 +24,7 @@ namespace TPP.Core.Modes
         private readonly ModeBase _modeBase;
         private readonly WebsocketBroadcastServer _broadcastServer;
         private readonly OverlayConnection _overlayConnection;
+        private readonly IBank<User> _pokeyenBank;
 
         public Matchmode(ILoggerFactory loggerFactory, BaseConfig baseConfig, MatchmodeConfig matchmodeConfig)
         {
@@ -29,7 +32,9 @@ namespace TPP.Core.Modes
             _loggerFactory = loggerFactory;
             _logger = loggerFactory.CreateLogger<Matchmode>();
             _stopToken = new StopToken();
-            _modeBase = new ModeBase(loggerFactory, baseConfig, _stopToken);
+            Setups.Databases repos = Setups.SetUpRepositories(baseConfig);
+            _pokeyenBank = repos.PokeyenBank;
+            _modeBase = new ModeBase(loggerFactory, repos, baseConfig, _stopToken);
 
             _broadcastServer = new WebsocketBroadcastServer(
                 loggerFactory.CreateLogger<WebsocketBroadcastServer>(), "localhost", 5001);
--- a/TPP.Core/Modes/Matchmode.cs
+++ b/TPP.Core/Modes/Matchmode.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading;
 using System.Threading.Tasks;
@@ -25,6 +26,7 @@ namespace TPP.Core.Modes
         private readonly WebsocketBroadcastServer _broadcastServer;
         private readonly OverlayConnection _overlayConnection;
         private readonly IBank<User> _pokeyenBank;
+        private readonly IUserRepo _userRepo;
 
         public Matchmode(ILoggerFactory loggerFactory, BaseConfig baseConfig, MatchmodeConfig matchmodeConfig)
         {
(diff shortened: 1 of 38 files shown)


Ravendwyr:
How possible is it to allow users to opt in to or out of the balance reset whispers?

Dnantz:
> How possible is it to allow users to opt in to or out of the balance reset whispers?

Seems extremely doable, just need to add a flag in the user's information and supporting commands.

TPP.Core/Modes/Matchmode.cs
```diff
@@ -132,5 +143,34 @@ public void Dispose()
...
+                    long amountToGive = subscriberMinimumPokeyen - pokeyen;
+                    await _pokeyenBank.PerformTransaction(new Transaction<User>(u, amountToGive, TPP.Persistence.Repos.TransactionType.Welfare));
+                    // TODO whisper users informing them they have been given money
+                    _logger.LogInformation(String.Format("Subscriber {0} had their balance reset to P{1} (+P{2})", u.SimpleName, subscriberMinimumPokeyen, amountToGive));
+
+                }
+                else if (!u.IsSubscribed && u.Pokeyen < minimumPokeyen)
+                {
+                    long amountToGive = minimumPokeyen - pokeyen;
+                    await _pokeyenBank.PerformTransaction(new Transaction<User>(u, amountToGive, TPP.Persistence.Repos.TransactionType.Welfare));
```
Felk, line 168: - `TPP.Persistence.Repos.TransactionType.Welfare` looks over-qualified
- the opportunity arises to use `PerformTransactions` (plural) to bundle all balance reset operations

TPP.Persistence/Repos/IUserRepo.cs
```diff
@@ -10,6 +11,7 @@ public interface IUserRepo
         public Task<User> RecordUser(UserInfo userInfo);
         public Task<User?> FindBySimpleName(string simpleName);
         public Task<User?> FindByDisplayName(string displayName);
+        public Task<List<User>> ListAllUnderPokeyen(long yen);
```
Felk, line 14: the method name seems odd compared to the rest. suggestion: `FindByPokeyenLessThan`

TPP.Core/Modes/Matchmode.cs
```diff
@@ -137,5 +143,34 @@ public void Dispose()
...
+                long pokeyen = await _pokeyenBank.GetAvailableMoney(u);
+                if(u.IsSubscribed && pokeyen < subscriberMinimumPokeyen)
+                {
+                    long amountToGive = subscriberMinimumPokeyen - pokeyen;
+                    await _pokeyenBank.PerformTransaction(new Transaction<User>(u, amountToGive, TPP.Persistence.Repos.TransactionType.Welfare));
+                    // TODO whisper users informing them they have been given money
+                    _logger.LogInformation(String.Format("Subscriber {0} had their balance reset to P{1} (+P{2})", u.SimpleName, subscriberMinimumPokeyen, amountToGive));
+
+                }
+                else if (!u.IsSubscribed && u.Pokeyen < minimumPokeyen)
```
Felk, line 165: You're using `u.Pokeyen` here, but the value returned by `await _pokeyenBank.GetAvailableMoney(u)` above. You don't need to use the available money, just the pokeyen amount on the user object should be fine, because
a) nothing except betting even reserves pokeyen (at this point in time)
b) if something were to exist to reserve pokeyen, it shouldn't cause you to get more welfare before a match

TPP.Core/Modes/Matchmode.cs
```diff
@@ -137,5 +143,34 @@ public void Dispose()
         {
             _modeBase.Dispose();
         }
+
+        public async void ResetBalances()
+        {
+            _logger.LogInformation("Resetting Balances");
```
Felk, line 149: feels like this should be debug level

TPP.Core/Modes/Matchmode.cs
```diff
@@ -137,5 +143,34 @@ public void Dispose()
         {
             _modeBase.Dispose();
         }
+
+        public async void ResetBalances()
```
Felk, line 147: never use `async void` (except for event handlers where not possible otherwise), always use `async Task`, and await it when called

Felk (changes requested):

Felk:
> > How possible is it to allow users to opt in to or out of the balance reset whispers?
>
> Seems extremely doable, just need to add a flag in the user's information and supporting commands.

Sure. If we were to do that, it should probably be part of a `Settings` dictionary to not introduce more stray fields on the user object

Felk:
requires https://github.com/TwitchPlaysPokemon/tpp-core/pull/210 to be merged first

TPP.Persistence/Repos/IUserRepo.cs
Dnantz: I originally named it like this to make it clear that it returns a list rather than a single user, however any IDE will tell you that anyway so I guess its rather unnecessary.

Felk: I believe "FindAll" is a good convention for this. That's how I've seen it in other ORMs at least

Felk:
oh, can you re-target the merge request to master please?

PR #215, opened 2021-03-26
Add role system

Dnantz:
Adds Roles for restricting command access.

Still to improve:
Write tests for new commands?
~~Automatically give broadcaster and console admin all permissions. Currently operators defined in the config bypass the permissions check, but this feels inelegant.~~
Config defined operators are now known as "default operators" and ideally should only include the streamer and essential staff. Old configs will need to be updated with the new syntax.

~~A slight difference from master:
When a user tries to use a command they don't have access to, the message has been changed from "only x can use this command" to "You do not have permission to use this command." Using the help command shows which group can use the command.~~
this has been reverted


Felk (changes requested):
there are a few fundamental things I want to discuss before going into a detailed review

TPP.Core/Commands/Definitions/OperatorCommands.cs
```diff
@@ -18,62 +19,88 @@ public class StopToken
...
+                Aliases = new [] { "trust", "trusted" },
+                Description = "Operator only: Give other users trusted status. Removes modteam roles. " +
+                              "Arguments: <user> <user> ..."
+            },
+            new Command("setmusicteam", setMusicTeam, UserGroup.Operator)
+            {
+                Aliases = new [] { "musicteam" },
+                Description = "Operator only: Give other users music team status. Removes modteam roles. " +
+                              "Arguments: <user> <user> ..."
+            },
```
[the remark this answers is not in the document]
Dnantz: I had set it up this way so that assigning one role would automatically remove other roles that weren't needed E.G. setting someone as moderator removes the now redundant trusted, musicteam roles in the case of a promotion. It also removes the operator role in the event of demotion, simplifying the entire interaction to one command call.

Doing the same in an [addrank <group>] style command I feel would require a less than ideal number of if statements and be messy.

I can change the output to show the modified user(s) updated ranks, however it would still require another call of removerank <group> to enforce users only having 1 of a set of hierarchic ranks.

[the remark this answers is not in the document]
Felk: I see what you mean. But we already have a metric shit ton of commands, so it feels like having the commands' implementation deal with a bit more stuff themselves in favor of only having to remember one commands feels preferable for me. Maybe other people in #new-core feel different about this

TPP.Core/Setups.cs
```diff
@@ -45,6 +45,7 @@ public static ArgsParser SetUpArgsParser(IUserRepo userRepo, PokedexData pokedex
             argsParser.AddArgumentParser(new ManyOfParser(argsParser));
 
             argsParser.AddArgumentParser(new UserParser(userRepo));
+            argsParser.AddArgumentParser(new RoleParser());
```
Felk, line 48: I failed to add comments to mark it, but I roughly divided this method into 3 sections of arguments parsers: "static" ones, self-referencial/recursive ones, and ones doing more fancy stuff with external dependencies. Can you push this line a few lines up, just below the pkmn species parser please?

TPP.ArgsParsing/TypeParsers/RoleParser.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    public class RoleParser : BaseArgumentParser<Role>
+    {
+        public override Task<ArgsParseResult<Role>> Parse(IImmutableList<string> args, Type[] genericTypes)
+        {
+            string roleToParse = args[0];
+            ArgsParseResult<Role> result;
+            try
+            {
+                Role parsedRole = (Role)Enum.Parse(typeof(Role), roleToParse, true);
```
Felk, line 22: Small QOL improvement so people don't have to check the documentation:
```suggestion
                Role parsedRole = (Role)Enum.Parse(typeof(Role), roleToParse, ignoreCase: true);
```

Felk (changes requested):
Looking good, on a conceptual level everything looks as good as it could be. I took a deeper dive into implementation details. Some nitpicks and a few possible code brevity improvements.

Regarding tests: I'd like to see at least one use case covered: Check that the role related commands do not work for non-OPs. The remaining doesn't seem all that critical that I wouldn't also be satisfied if it was just tested by hand once.

Felk:
Oh and make sure to fix the conflicts so the automated pipelines can run. There's a few formatting issues I didn't bother pointing out by hand

Felk:
another feature request: can you add a command that lists all users that have a specific role?

Felk:
noticed another thing: please also rename the OperatorNames config in TPP.Core's README to reflect the new one

Dnantz:
> another feature request: can you add a command that lists all users that have a specific role?

would you rather have individual commands like the old core (ops, mods, ...) or a single !list <role>

TPP.Core/Commands/Definitions/OperatorCommands.cs
```diff
@@ -224,26 +218,18 @@ public async Task<CommandResult> AddRole(CommandContext context)
             (User user, Role role) = await context.ParseArgs<User, Role>();
             string response;
 
-            if (user.Roles != null)
+            HashSet<Role> roles = user.Roles;
+            bool roleAssigned = user.Roles.Add(role);
```
Felk, line 222: I think you misunderstood me. Please don't modify the Roles set of user instances directly. I think it's fine to keep the actual type a HashSet for the reason that mongodb doesn't make it easy to use an actually immutable set, but the user object should still be treated immutable. Same goes for the Remove call further down.

It doesn't make much of a difference for this case, so it's not a big deal. If there's a good reason to mutate it in-place, you may keep it

TPP.Core/Commands/Definitions/UserCommands.cs
```diff
@@ -59,6 +59,17 @@ public class UserCommands : ICommandCollection
...
+            {
+                Description = "List all of the people who have a given role." +
+                              "Arguemnts: <Role>"
+            },
+            new Command("showroles", ShowRoles)
+            {
+                Aliases = new[] {"roles"},
+                Description = "Operators only: Show which roles a user has." +
+                              "Arguments: <user>"
+            }
```
Felk, line 76: pls use trailing commas where possible to improve git diffs when people add further entries. Same goes for the list of operator commands

TPP.Core/Commands/Definitions/UserCommands.cs
```diff
@@ -228,5 +239,25 @@ public async Task<CommandResult> Donate(CommandContext context)
...
         }
+
+        public async Task<CommandResult> List(CommandContext context)
+        {
+            Role role = await context.ParseArgs<Role>();
+            List<User> users = await _userRepo.FindAllByRole(role);
+
+            return new CommandResult
+            {
+                Response = users.Count > 0 ? $"The users with the '{role.ToString()}' role are: {string.Join(", ", users.Select(u => u.SimpleName))}" : $"There are no users with the '{role.ToString()}' role."
```
Felk, line 277: can you split the ternary into multiple lines for better readability? Same for the ShowRoles command

Felk: Also please use Name instead of SimpleName for all user-facing texts. logging entries should just rely on User objects' ToString method since that also prints the user ID which can be useful. I don't know of any use cases where outputting the SimpleName is preferable

TPP.Persistence.MongoDB/Repos/UserRepo.cs
```diff
@@ -44,17 +47,26 @@ static UserRepo()
                 cm.MapProperty(u => u.GlowColor).SetElementName("secondary_color");
                 cm.MapProperty(u => u.GlowColorUnlocked).SetElementName("secondary_color_unlocked");
                 cm.MapProperty(u => u.PokeyenBetRank).SetElementName("pokeyen_bet_rank");
-                cm.MapProperty(u => u.Roles).SetElementName("roles");
+                cm.MapProperty(u => u.Roles).SetElementName("roles")
+                .SetDefaultValue(new HashSet<Role>());
```
Felk, line 61: indentation looks wrong

TPP.Persistence.MongoDB/Repos/UserRepo.cs
```diff
@@ -44,17 +47,26 @@ static UserRepo()
...
             _startingPokeyen = startingPokeyen;
             _startingTokens = startingTokens;
+            _defaultOperators = defaultOperators.ToImmutableHashSet<string>();
             InitIndexes();
+
+            foreach (string name in _defaultOperators)
+            {
+                User? user = FindBySimpleName(name.ToLower()).Result;
+                if (user != null)
+                    SetRoles(user, new HashSet<Role> { Role.Operator });
```
Felk, line 78: this looks like it also removes any existing roles

TPP.Persistence.MongoDB/Repos/UserRepo.cs
```diff
@@ -123,6 +135,8 @@ public async Task<User> RecordUser(UserInfo userInfo)
             try
             {
                 await Collection.InsertOneAsync(document: user);
+                if (_defaultOperators.Contains(user.SimpleName))
+                    await SetRoles(user, new HashSet<Role> { Role.Operator });
```
Felk, line 147: you can already specify the roles in the User object's constructor instead of adding them here retroactively

Felk (changes requested):

TPP.Core/Commands/Definitions/OperatorCommands.cs
Dnantz: This was a typo, I meant to modify the copy "roles" just above but... didn't.

Felk: the line above it isn't a copy though

Dnantz: you have a point. That will be changed.

TPP.Persistence.MongoDB/Repos/UserRepo.cs
Dnantz: It was written with the assumption that operator is a sort of "master role" and wouldn't need (or make sense to have) other roles. This can be changed.

Felk: hmm you have a point. I'm indifferent about this now

Felk (approved):

PR #217, opened 2021-03-31
Poll feature (continuation)

Felk:


chaos-lord (changes requested):
Voting for 0 options is possible and locks you out of voting again, either add the ability to change votes or cause 0 response answers to error

chaos-lord (approved):
Yep, seems to do what we need for now, we will need to add a !closepoll at some point but that's not blocking.

TPP.ArgsParsing/TypeParsers/BoolParser.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading.Tasks;
+
+namespace TPP.ArgsParsing.TypeParsers
+{
+    public class BoolParser : BaseArgumentParser<bool>
+    {
+        public override Task<ArgsParseResult<bool>> Parse(IImmutableList<string> args, Type[] genericTypes) =>
+            Task.FromResult(args[0].ToLower() switch
+            {
+                "yes" or "true" => ArgsParseResult<bool>.Success(true, args.Skip(1).ToImmutableList()),
```
red031000, line 13: y/n might be a good idea to accept as well

TPP.Core/Commands/CommandExtensions.cs
```diff
@@ -19,6 +19,9 @@ public static class CommandExtensions
         public static Task<(T1, T2, T3, T4)> ParseArgs<T1, T2, T3, T4>(this CommandContext ctx) =>
             ctx.ArgsParser.Parse<T1, T2, T3, T4>(ctx.Args);
 
+        public static Task<(T1, T2, T3, T4, T5)> ParseArgs<T1, T2, T3, T4, T5>(this CommandContext ctx) =>
+            ctx.ArgsParser.Parse<T1, T2, T3, T4, T5>(ctx.Args);
+
         /// Replace the command execution with a different one.
         public static Command WithExecution(this Command command, Command.Execute newExecution)
         {
```
red031000, line 27: please clean up warning below

TPP.Core/Commands/Definitions/ManagePollCommands.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public ManagePollCommands(IPollRepo pollRepo)
+        {
+            _pollRepo = pollRepo;
+        }
+
+        public async Task<CommandResult> StartPoll(CommandContext context)
+        {
+            (string pollName, string pollCode, bool multiChoice, bool allowChangeVote, ManyOf<string> options) =
+                await context.ParseArgs<string, string, bool, bool, ManyOf<string>>();
```
red031000, line 37: imo multiChoice and allowChangeVote should be optional and should have a default value of false, if the bool parser can't parse params 3 and 4 into a bool they should be grouped into options, may require a change to args parser for that though

TPP.Core/Commands/Definitions/ManagePollCommands.cs
```diff
@@ -0,0 +1,54 @@
...
+            await _pollRepo.CreatePoll(pollCode, pollName, multiChoice, allowChangeVote, options.Values);
+            return new CommandResult
+            {
+                Response =
+                    $"Poll '{pollCode}' created: {pollName}" +
+                    $" - {(multiChoice ? "multiple-choice" : "single-choice")}" +
+                    $" - {(allowChangeVote ? "changeable votes" : "unchangeable votes")}"
+            };
+        }
+    }
```
red031000, line 53: please add RemovePoll/DeletePoll as a matter of urgency, this should not have been shipped without a way to remove the created poll

TPP.Core/Commands/Definitions/PollCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            Poll? poll = await _pollRepo.FindPoll(pollCode);
+            if (poll == null)
+                return new CommandResult { Response = $"No poll with the code '{pollCode}' was found." };
+
+            List<int> selectedOptions = new();
+            foreach (string voteStr in votes)
+            {
+                PollOption? option = null;
+                if (int.TryParse(voteStr.TrimStart('#'), out int voteInt))
```
red031000, line 53: letters (i.e. ABCD) should be valid options imo, though should be regulated in the extremely unlikely/impossible event that there are more than 26 options

TPP.Core/Commands/Definitions/ManagePollCommands.cs
```diff
@@ -0,0 +1,54 @@
...
+        public async Task<CommandResult> StartPoll(CommandContext context)
+        {
+            (string pollName, string pollCode, bool multiChoice, bool allowChangeVote, ManyOf<string> options) =
+                await context.ParseArgs<string, string, bool, bool, ManyOf<string>>();
+            if (options.Values.Count < 2) return new CommandResult { Response = "must specify at least 2 options" };
+            pollName = pollName.Replace('_', ' ');
+
+            if (await _pollRepo.FindPoll(pollCode) != null)
+                return new CommandResult { Response = $"A poll with the code '{pollCode}' already exists." };
+
```
red031000, line 43: imo there should be a check if two of the options are the same, and fail the poll if they are

TPP.Core/Commands/Definitions/PollCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+                    option = poll.PollOptions.FirstOrDefault(o => o.Id == voteInt);
+                option ??= poll.PollOptions.FirstOrDefault(o => o.Option == voteStr);
```
red031000, line 55: this should be case insensitive

TPP.Core/Commands/Definitions/PollCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+        public async Task<CommandResult> Poll(CommandContext context)
+        {
+            string pollCode = await context.ParseArgs<string>();
+            Poll? poll = await _pollRepo.FindPoll(pollCode);
+            if (poll == null)
+                return new CommandResult { Response = $"No poll with the code '{pollCode}' was found." };
+
+            string Percentage(PollOption option) => poll.Voters.Count == 0
+                ? "0" // avoid division by zero
+                :$"{100 * (option.VoterIds.Count / (double) poll.Voters.Count):0.#}";
```
red031000, line 91: two errors on this line, please fix

TPP.Core/Commands/Definitions/PollCommands.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            return new CommandResult
+            {
+                Response = $"Poll '{poll.PollCode}': {poll.PollTitle} - " + string.Join(", ",
+                    poll.PollOptions.Select(option => $"#{option.Id} {option.Option} ({Percentage(option)}%)"))
```
red031000, line 96: vote numbers should be included here

TPP.Persistence.MongoDB/Repos/PollRepo.cs
```diff
@@ -0,0 +1,132 @@
...
+            if (options.Count == 0)
+                return new VoteFailure.CannotVoteForNone();
+            if (options.Count > 1 && !poll.MultiChoice)
+                return new VoteFailure.NotMultipleChoice();
+
+            await Collection.UpdateOneAsync(
+                p => p.PollCode == pollCode,
+                Builders<Poll>.Update.AddToSet(p => p.Voters, userId));
+
+            ImmutableList<int> invalidOptions = options.Except(poll.PollOptions.Select(p => p.Id)).ToImmutableList();
```
red031000, line 105: invalid options should be checked before the voter is added to the db, otherwise this breaks AllowChangeVote being set to false

TPP.Persistence.MongoDB/Repos/PollRepo.cs
```diff
@@ -0,0 +1,132 @@
...
+            if (invalidOptions.Any())
+                return new VoteFailure.InvalidOptions(invalidOptions);
+
+            // remove any existing votes in case the user has voted before and is changing their vote right now.
+            // no typed support for the $[] (for each in array) operator yet, see https://jira.mongodb.org/browse/CSHARP-2232
+            foreach (int i in Range(0, poll.PollOptions.Count))
```
red031000, line 111: should only be done if AllowChangeVote is true, otherwise is a waste of time

TPP.Core/Commands/Definitions/ManagePollCommands.cs
```diff
@@ -0,0 +1,54 @@
...
+        public async Task<CommandResult> StartPoll(CommandContext context)
+        {
+            (string pollName, string pollCode, bool multiChoice, bool allowChangeVote, ManyOf<string> options) =
+                await context.ParseArgs<string, string, bool, bool, ManyOf<string>>();
+            if (options.Values.Count < 2) return new CommandResult { Response = "must specify at least 2 options" };
+            pollName = pollName.Replace('_', ' ');
+
+            if (await _pollRepo.FindPoll(pollCode) != null)
+                return new CommandResult { Response = $"A poll with the code '{pollCode}' already exists." };
```
red031000, line 42: this should only be returned if the found poll is marked as active, otherwise imo it's ok to pose the same question twice, using the most recent as the one returned etc

red031000 (commented):
I can't request changes here but please implement these

TPP.Core/Commands/CommandExtensions.cs
Felk: what warnings specifically?

TPP.Core/Commands/Definitions/ManagePollCommands.cs
Felk: sounds unnecessarily complicated to me since creating polls is a rare and administrative process anyway

TPP.Core/Commands/Definitions/PollCommands.cs
Felk: I don't see errors, but I suppose you mean non-conforming formatting

TPP.Core/Commands/CommandExtensions.cs
red031000:
```
TPP.Core/Commands/CommandExtensions.cs#L50
Parameter 'ctx' is never used
```

TPP.Core/Commands/Definitions/ManagePollCommands.cs
red031000: that's fair I guess, not really necessary

TPP.Core/Commands/Definitions/PollCommands.cs
red031000: both errors are visible here
https://github.com/TwitchPlaysPokemon/tpp-core/pull/217/files#diff-ad784caac43714628eca51b1aaf20f75f3a6242c547fafd2429cd3867a694906R91

TPP.Persistence.MongoDB/Repos/PollRepo.cs
Felk: I'd rather keep it for robustness. This code is by no means latency critical

PR #223, opened 2021-04-01
Pokedex extensions

Fenris2142:
Still a WIP, but I got an exception when trying out some new stuff I've been working on. It happens when PkmnSpeciesExtensions.GetGeneration is called with one of the old Fakemons (pre Sirius era). Thought I might aswell make the fix a seperate commit in case someone else is working with the function right now.

Obviously this is just a temporary fix and can be removed again as soon as the old Fakemons got proper Fakemon ID's

image


Felk:
hmm... I don't think we should ever change any species ID. I fear too much data has escaped TPP's database and lives in the outside world now. Can you change this to be a permanent rule of everything >= 2000 being consider a fakemon? And stuff below that still being invalid

Felk:
You may instead add a temporary fix to the "IsFakemon"-property of `PkmnSpecies` instead, which will then automatically also work here

Fenris2142:
> You may instead add a temporary fix to the "IsFakemon"-property of `PkmnSpecies` instead, which will then automatically also work here

done

PR #224, opened 2021-04-02
Support for Regions (Generations) for the pokedex command.

Fenris2142:
See image for examples. I added support for all 8 regions, also one entry for fakemons and a special easter egg, which surely won't get removed, right? 🦘 🐕 

image


TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -253,18 +267,51 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+                ImmutableSortedDictionary<PkmnSpecies, int> ownedPokemons = (await _badgeRepo.CountByUserPerSpecies(user.Id));
+                int userOwnedRegionCount = ownedPokemons.Count(ownedPokemon =>
+                    ownedPokemon.Key.GetGeneration() == modeGeneration);
+                return new CommandResult
+                {
+                    Response = isSelf
+                        ? $"You have collected {userOwnedRegionCount}/{totalRegionCount} "
+                          + $"distinct {System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(mode)} badge(s)"
+                        : $"{user.Name} has collected {userOwnedRegionCount}/{totalRegionCount} "
+                          + $"distinct {System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(mode)} badge(s)"
```
Felk, line 298: does that really need to be qualified with `System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo`? Also, is there a culture-independent way of doing this? Is just uppercasing the first letter sufficient? E.g. `mode[0].ToUpper() + mode[1..]`

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -253,18 +267,51 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+            {
+                ImmutableSortedDictionary<PkmnSpecies, int> ownedPokemons = (await _badgeRepo.CountByUserPerSpecies(user.Id));
+                int userOwnedRegionCount = ownedPokemons.Count(ownedPokemon =>
+                    ownedPokemon.Key.GetGeneration() == Generation.Gen1);
+                return new CommandResult
+                {
+                    Response = isSelf
+                        ? $"You have collected {userOwnedRegionCount}/151 distinct KANTOOOOOOOOOOOOOOO OhMyDog badge(s)"
+                        : $"{user.Name} has collected {userOwnedRegionCount}/151 distinct KANTOOOOOOOOOOOOOOO OhMyDog badge(s)"
+                };
```
Felk, line 280: could this be done without duplicating code of the other mode branch?

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -253,18 +267,51 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+            }
+            else if (_pokedexModeRegions.ContainsKey(mode.ToLower()))
```
Felk, line 282: just lowercasing the mode once at the beginning of the entire pokdex command would also benefit the other modes and avoid duplicating calls to `ToLower()`

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -253,18 +267,51 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
-                        + $"'{PokedexModeComplementFromDupes}' Compares missing badges from User A with owned duplicate badges from User B"
+                        + $"'{PokedexModeComplementFromDupes}' Compares missing badges from User A with owned duplicate badges from User B, "
+                        + $"'{_pokedexModeRegions.ElementAt(0).Key}','{_pokedexModeRegions.ElementAt(1).Key}',... Shows how many badges you or the specified user owns from the specified region only"
                 };
             }
             return new CommandResult
             {
-                Response = $"Unsupported mode '{mode}'. Current modes supported: {PokedexModeModes}, {PokedexModeDupes}, {PokedexModeMissing}, {PokedexModeComplementFromDupes}, {PokedexModeComplementFrom}"
+                Response = $"Unsupported mode '{mode}'. Current modes supported: {PokedexModeModes}, {PokedexModeDupes}, {PokedexModeMissing}, {PokedexModeComplementFromDupes}, "
+                    + $"{PokedexModeComplementFrom}, {_pokedexModeRegions.ElementAt(0).Key}, {_pokedexModeRegions.ElementAt(1).Key},... "
```
Felk, line 314: I believe using `ElementAt` with a non-sorted dictionary will cause this command to effectively print two random region names. The same goes for other occurances of `ElementAt`

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -253,18 +267,51 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+                        : $"{user.Name} has collected {userOwnedRegionCount}/151 distinct KANTOOOOOOOOOOOOOOO OhMyDog badge(s)"
+                };
+            }
+            else if (_pokedexModeRegions.ContainsKey(mode.ToLower()))
+            {
+                Generation modeGeneration = _pokedexModeRegions[mode.ToLower()];
+
+                // Getting the total number of distinct Pokemon from a single generation could also be a function in class PokedexData,
+                //   by using "GenMaxIds", but I'm not sure how to get the total number of Fakemons
+                int totalRegionCount = _knownSpecies.Count(pokemon => pokemon.GetGeneration() == modeGeneration);
```
Felk, line 288: This could be done once in the constructor for all generations instead of each time the command runs

Felk (changes requested):

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: Well, just **TextInfo.ToTitleCase** doesn't work as stated here

https://docs.microsoft.com/en-us/dotnet/api/system.globalization.textinfo.totitlecase?view=net-5.0

> culture-independent
what does that mean. Is using CurrentCulture bad?

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: I guess the attempt to make this as generic as possible was too sophisticated anyway. The chances that a region gets renamed are near 0, so I might as well just add them as hardcoded examples here.

TPP.Core/Commands/Definitions/BadgeCommands.cs
Fenris2142: Removed the code for now, but I'll try to make this one work when I find the motivation

>don't make the easter egg mandatory. I believe some people might actually still parse bot output instead of using the API
> you can try to make the last letter for all regions repeatable
> johtooo -> JOHTOOO :dog:

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -10,6 +10,20 @@
...
+    public class RegionInformation
+    {
+        public Generation Generation { get; }
+        public int TotalRegionCount { get; }
+        public RegionInformation(Generation generation, int totalRegionCount)
+        {
+            Generation = generation;
+            TotalRegionCount = totalRegionCount;
+        }
+    }
```
Felk, line 25: You may simply use a record for this:

```suggestion
    public record RegionInformation(Generation Generation, int TotalRegionCount);
```

PR #228, opened 2021-04-19
Pokedex extensions

Fenris2142:
I kinda forgot about "Natdex", when I added the regions. Chat requested the feature, so I thought: Might aswell implement it. Also it was mentioned here:

[https://github.com/TwitchPlaysPokemon/public/issues/129](url)

, so I guess this issue can be closed after the PR is finished. 

Commit should be easy enough to understand. "National"-Dex is just the inverse of Fakemons


TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -275,8 +277,17 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+                if (mode.Equals(PokedexModeNational))
+                {
+                    userOwnedRegionCount = ownedPokemons.Count(ownedPokemon =>
+                        ownedPokemon.Key.GetGeneration() != regionInformation.Generation);
+                }
+                else
+                {
+                    userOwnedRegionCount = ownedPokemons.Count(ownedPokemon =>
+                        ownedPokemon.Key.GetGeneration() == regionInformation.Generation);
+                }
```
Felk, line 290: sure, but can you make that more compact?
```suggestion
                int userOwnedRegionCount = mode.Equals(PokedexModeNational)
                    ? ownedPokemons.Count(ownedPokemon => ownedPokemon.Key.GetGeneration() != regionInformation.Generation)
                    : ownedPokemons.Count(ownedPokemon => ownedPokemon.Key.GetGeneration() == regionInformation.Generation);
```

Felk (approved):

PR #229, opened 2021-04-19
rework formatting workflow, add resharper analyze github workflow

Felk:
this combines the previous formatting workflow with additional resharper inspections into a new "inspect" workflow.


Felk:
alright, that should do it

red031000 (approved):

PR #231, opened 2021-04-26
Badge trading

Mogiiii:
WIP pull request for badge trading and updates to badges in general.


TPP.ArgsParsing/TypeParsers/ShinyParser.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using TPP.Common;
+
+namespace TPP.ArgsParsing.TypeParsers
+{
+    /// <summary>
+    /// A parser that finds a badge form by name.
+    /// </summary>
+    public class ShinyParser : BaseArgumentParser<bool>
```
Felk, line 12: this clashes with the generic `BoolParser` that has been added on the master branch by now. You need to wrap the enum in a dedicated type to be able to use it for the args parser. You can look at `Pokeyen` for example to see what I mean.

TPP.Persistence.MongoDB/Repos/BadgeRepo.cs
```diff
@@ -41,6 +41,8 @@ static BadgeRepo()
                     .SetIgnoreIfNull(true);
                 cm.MapProperty(b => b.SellingSince).SetElementName("selling_since")
                     .SetIgnoreIfNull(true);
+                cm.MapProperty(b => b.Shiny).SetElementName("shiny")
+                    .SetDefaultValue(false);
```
Felk, line 45: If we deliberately have non-shinies not have the field in the database, this needs to also have `.SetIgnoreIfDefault` (or similar), so it doesn't become a mixed bag between "shiny: false" and "missing 'shiny' field"

TPP.Persistence.MongoDB/Repos/BadgeRepo.cs
```diff
@@ -102,6 +103,7 @@ public async Task<List<Badge>> FindAllByCustom(string? userId = null, PkmnSpecie
                 filter &= Builders<Badge>.Filter.Eq(b => b.Form, form);
             if (source != null)
                 filter &= Builders<Badge>.Filter.Eq(b => b.Source, source);
+            filter &= Builders<Badge>.Filter.Eq(b => b.Shiny, shiny); // always assigned, defaults to false
```
Felk, line 106: Hmm, maybe queries do become awkward at places without a migration after all. Does this work properly for `false`? Because it might not, since null != false :/

TPP.Core/Commands/Definitions/OperatorCommands.cs
```diff
@@ -184,11 +184,11 @@ public async Task<CommandResult> TransferBadge(CommandContext context)
...
         public async Task<CommandResult> CreateBadge(CommandContext context)
         {
-            (User recipient, PkmnSpecies species, Optional<PositiveInt> amountOpt, Optional<string> formOpt, Optional<bool> shinyOpt) =
-                await context.ParseArgs<AnyOrder<User, PkmnSpecies, Optional<PositiveInt>, Optional<string>, Optional<bool>>>();
+            (User recipient, PkmnSpecies species, Optional<PositiveInt> amountOpt, Optional<string> formOpt, Optional<Shiny> shinyOpt) =
+                await context.ParseArgs<AnyOrder<User, PkmnSpecies, Optional<PositiveInt>, Optional<string>, Optional<Shiny>>>();
             int amount = amountOpt.Map(i => i.Number).OrElse(1);
             int form = PkmnForms.pokemonHasForms(species) ? 0 : 1; // default to the first listed form if form is unspecified
-            bool shiny = shinyOpt.IsPresent ? shinyOpt.Value : false;
+            bool shiny = shinyOpt.Value ?? false;
```
Felk, line 191: Does `??` work with the optional omitted? Accessing `.Value` throws an exception if the optional is empty

TPP.Persistence.MongoDB/Repos/BadgeRepo.cs
Dnantz: assuming this works the way I think it does

cm.MapProperty(b => b.Shiny).SetElementName("shiny")
                    .SetDefaultValue(false)
                    .SetIgnoreIfDefault(true);

when it loads things from the database it should interpret missing fields and null values as false.

Felk: yeah but this is a filter for a query and it might get serialized to mongodb syntax as-is. Can you write a test to verify this works for nulls? You can look at e.g. `BankTest#can_handle_null_transaction_type` to see how to insert raw bson data in a test.
If you don't have the time or don't wanna do it for any reason let me know, then I can test it.

Dnantz: I should be able to get to it soon™

TPP.Persistence.MongoDB.Tests/Repos/BadgeRepoTest.cs
```diff
@@ -190,6 +197,33 @@ public async Task FindAllForSaleByCustom_only_returns_badges_for_sale()
...
+                ["source"] = "manual_creation",
+                ["created_at"] = instant.ToDateTimeUtc(),
+                ["form"] = 0,
+            }));
+
+            IMongoCollection<Badge> badgeCollection = db.GetCollection<Badge>("badges"); ;
+
+            Badge b = await badgeCollection.Find(b => b.Id == id).FirstAsync();
+
+            Assert.AreEqual(false, b.Shiny);
```
Felk, line 224: This is a good test to see that an absent "shiny" field gets properly deserialized as "false" in the model class. Can you add an additional test case that calls `FindAllByCustom` with shiny=false to see whether that filter will also find badges with an unset or null shiny field?

TPP.Persistence.MongoDB/Repos/BadgeRepo.cs
```diff
@@ -103,8 +103,10 @@ public async Task<List<Badge>> FindAllByCustom(string? userId = null, PkmnSpecie
                 filter &= Builders<Badge>.Filter.Eq(b => b.Form, form);
             if (source != null)
                 filter &= Builders<Badge>.Filter.Eq(b => b.Source, source);
-            if (shiny != null)
-                filter &= Builders<Badge>.Filter.Eq(b => b.Shiny, shiny);
+            if (shiny == true)
+                filter &= Builders<Badge>.Filter.Eq(b => b.Shiny, true);
+            else
+                filter &= Builders<Badge>.Filter.Ne(b => b.Shiny, true);
```
Felk, line 109: Pls add a short comment along the lines of `cannot test for == false, because that would not match null`

TPP.Persistence.MongoDB/Repos/BadgeRepo.cs
```diff
@@ -64,26 +68,55 @@ private void InitIndexes()
         }
```
Felk, line 68: (above here, can't comment on the correct line).
Please add indexes for the new filterable fields. 
Nitpick: Also this is fine for now, but I suspect filtering for "!= true" for the shiny field _might_ not work properly for the index. Maybe it does, but it's something you can test if you're still motivated. E.g. by enabling [profiling level 2](https://docs.mongodb.com/manual/tutorial/manage-the-database-profiler/), issuing a query, and looking at the execution plan.

TPP.Core/Commands/Definitions/OperatorCommands.cs
```diff
@@ -183,18 +184,38 @@ public async Task<CommandResult> TransferBadge(CommandContext context)
 
         public async Task<CommandResult> CreateBadge(CommandContext context)
         {
-            (User recipient, PkmnSpecies species, Optional<PositiveInt> amountOpt) =
-                await context.ParseArgs<AnyOrder<User, PkmnSpecies, Optional<PositiveInt>>>();
+            (User recipient, PkmnSpecies species, Optional<PositiveInt> amountOpt, Optional<string> formOpt, Optional<Shiny> shinyOpt) =
+                await context.ParseArgs<AnyOrder<User, PkmnSpecies, Optional<PositiveInt>, Optional<string>, Optional<Shiny>>>();
             int amount = amountOpt.Map(i => i.Number).OrElse(1);
-
+            int form = PkmnForms.pokemonHasForms(species) ? 0 : 1; // default to the first listed form if form is unspecified
```
Felk, line 190: I don't understand the comment. Why `1`? Isn't `0` an acceptable default for both cases?

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -300,5 +324,62 @@ public async Task<CommandResult> GiftBadge(CommandContext context)
...
+                response = string.Format("{0} badges found:", forSale.Count);
+                foreach (Badge b in forSale)
+                {
+                    if (b.UserId == null)
+                        throw new OwnedBadgeNotFoundException(b);
+                    User? seller = await _userRepo.FindById(b.UserId);
+                    if (seller == null)
+                        throw new OwnedBadgeNotFoundException(b);
+
+                    response += string.Format("\n{0}{1}{2} sold by {3} for T{4}",
```
Felk, line 370: can't have newlines in IRC messages

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -271,15 +277,33 @@ public async Task<CommandResult> Pokedex(CommandContext context)
...
+                    form = PkmnForms.getFormId(species, formName);
+                }
+                catch (ArgumentException e)
+                {
+                    return new CommandResult
+                    {
+                        Response = e.Message
+                    };
+                }
+            }
```
Felk, line 300: This large block of code is duplicated a few times. You can either extract it and/or reduce its size. I personally find the following quite aesthetically acceptable:
```c#
int? form;
try {
    form = formOpt
        .Map(name => PkmnForms.getFormId(species, name))
        .OrElse(null);
} catch (ArgumentException e) {
    return new CommandResult { Response = e.Message };
}
```

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -48,8 +49,13 @@ public class BadgeCommands : ICommandCollection
             new Command("giftbadge", GiftBadge)
             {
                 Description =
-                    "Gift a badge you own to another user with no price. Arguments: <pokemon> <number of badges>(Optional) <username>"
+                    "Gift a badge you own to another user with no price. Arguments: <pokemon> <number of badges>(Optional) <username>" //TODO also update this before merge
             },
+            new Command("listsellbadge", ListSellBadge)
+            {
+                Description =
+                    "List the badges someone is selling. Arguments: <pokemon> <username> (optional) <shiny> <source> (optional) (optional) <form> (optional)"
```
Felk, line 57: `<shiny> <source> (optional) (optional)` looks wrong

TPP.Core.Tests/Commands/Definitions/BadgeCommandsTest.cs
```diff
@@ -306,5 +306,27 @@ public async Task TestGiftBadgeNotOwned()
...
+
+            _badgeRepoMock.Setup(repo => repo.FindAllForSaleByCustom(user1.Id, speciesA, null, null, false)).Returns(Task.FromResult(new List<Badge>() { badgeA }));
+            _userRepoMock.Setup(repo => repo.FindById(user1.Id)).Returns(Task.FromResult((User?)user1));
+
+            _argsParser.AddArgumentParser(new PkmnSpeciesParser(new[] { speciesA }));
+            _argsParser.AddArgumentParser(new ShinyParser());
+            _argsParser.AddArgumentParser(new BadgeSourceParser());
+            _argsParser.AddArgumentParser(new StringParser());
+
+            await _badgeRepoMock.Object.SetBadgeSellPrice(badgeA, 1);
```
Felk, line 325: this line probably has no effect. You're calling a method on a mock. The return value of `FindAllForSaleByCustom` is already mocked above

TPP.Common/PkmnForms.cs
```diff
@@ -0,0 +1,63 @@
...
+            return formName;
+        }
+
+        public static int getFormId(PkmnSpecies pokemon, string formName)
+        {
+            string pkmnName = pokemon.Name.ToLower();
+            Dictionary<string, int>? forms;
+            if (!Forms.TryGetValue(pkmnName, out forms))
+                throw new ArgumentException($"{pokemon.Name} does not have alternate forms.");
+            int formid = forms.GetValueOrDefault(formName.ToLower());
```
Felk, line 52: Looks like you don't need the `OrDefault`, since you already checked that it's present

TPP.ArgsParsing/TypeParsers/ShinyParser.cs
```diff
@@ -0,0 +1,45 @@
...
+        };
+        string[] plainWords =
+        {
+            "plain",
+            "regular",
+            "shiny:false"
+        };
+        public override Task<ArgsParseResult<Shiny>> Parse(IImmutableList<string> args, Type[] genericTypes)
+        {
+            string s = args[0];
```
Felk, line 28: probably want to make that lowercase

Felk (changes requested):
Very neat so far! You have written few TODOs, but I also don't mind merging an intermediate state as long the commands aren't hooked up yet and therefore aren't part of production code

TPP.Core/Commands/Definitions/OperatorCommands.cs
Dnantz: Forms are 1-indexed. The idea is that 0 means forms don't exist for the species, and positive ints are used for the forms of species that have forms. E.G. a form 0 shellos is invalid, and a form 1 caterpie is also invalid.

Felk: according to this table: https://pixelmonmod.com/wiki/Form_indices they are 0-indexed, and shellos 0 is shellos east for example

Felk: I think it's beneficial to have the form numbers we use be completely synchronized with the official form indexes, so we don't need to remember to apply some mapping every time we generate a badge with a form number

Dnantz: I wasn't aware this existed. If this matches the official specification I'll just use the values here.

Felk: no idea if that website is "official" in all cases, but official form numbers do exist. I couldn't find a more "official" list

TPP.Core/Commands/Definitions/BadgeCommands.cs
Dnantz: This was originally much worse, and will almost certainly be moved to a method along with other related badge info.

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -93,6 +99,17 @@ public class BadgeCommands : ICommandCollection
...
 
+        /// <summary>
+        /// Convineintly handles many optional arguments for badge commands. May throw an exception when given faulty form informaton, so this should be surrounded by a try/catch
+        /// </summary>
+        private (Badge.BadgeSource?, string?, bool?) InterpretBadgeInfoArgs(Optional<Badge.BadgeSource> sourceOpt, Optional<Form> formOpt, Optional<Shiny> shinyOpt)
+        {
+            Badge.BadgeSource? source = sourceOpt.IsPresent ? sourceOpt.Value : null;
+            string? form = formOpt.IsPresent ? formOpt.Value.Name : null;
+            bool? shiny = shinyOpt.IsPresent ? shinyOpt.Value : false;
+            return (source, form, shiny);
```
Felk, line 110: This looks equivalent to
```
return (sourceOpt.OrElse(null), formOpt.OrElse(null), shinyOpt.OrElse(false));
```

TPP.Core/Commands/Definitions/BadgeCommands.cs
```diff
@@ -339,5 +358,44 @@ public async Task<CommandResult> GiftBadge(CommandContext context)
...
+                    User? seller = await _userRepo.FindById(b.UserId);
+                    if (seller == null)
+                        throw new OwnedBadgeNotFoundException(b);
+
+                    response += string.Format(" {0}{1}{2} sold by {3} for T{4}",
+                        shiny == true ? "Shiny " : "",
+                        form ?? "",
+                        b.Species.Name,
+                        seller.SimpleName,
+                        b.SellPrice);
```
Felk, line 391: avoid building strings in a loop with `+=`. Instead, Build an enumerable and `string.Join` it.

TPP.Core/Commands/Definitions/BadgeCommands.cs
Felk: also is defaulting to shiny=false correct? It's still nullable, which indicates to me you wanted it to be kind of a tri-state of "yes", "no" and "don't care whether shiny or not"

TPP.Core/Setups.cs
```diff
@@ -42,6 +42,9 @@ public static ArgsParser SetUpArgsParser(IUserRepo userRepo, PokedexData pokedex
             argsParser.AddArgumentParser(new TokensParser());
             argsParser.AddArgumentParser(new SignedPokeyenParser());
             argsParser.AddArgumentParser(new SignedTokensParser());
+            argsParser.AddArgumentParser(new PkmnSpeciesParser(pokedexData.KnownSpecies, PokedexData.NormalizeName));
```
Felk, line 45: this looks duplicated now

TPP.Model/BadgeBuyOffer.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// When the buy offer was created.
+        /// </summary>
+        public Instant CreatedAt { get; init; }
+
+        /// <summary>
+        /// When this offer was last updated.
+        /// </summary>
+        public Instant WaitingSince { get; private set; }
```
Felk, line 61: xml doc and property name seem in conflict. If it's better described as "last updated" it should be called `LastUpdatedAt` and the redundant documentation maybe even removed.

TPP.Model/BadgeBuyOffer.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        //duration and expires_at depricated from old core
```
Felk, line 63: typo: depricated -> deprecated

TPP.Model/BadgeBuyOffer.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// Decrement the amount to buy.
+        /// </summary>
+        public void decrement(Instant decrementedAt)
+        {
+            if (Amount <= 0)
+                throw new InvalidOperationException("The buy offer has no badges remaining, and cannot be decremented further.");
+            Amount--;
+            WaitingSince = decrementedAt;
+        }
```
Felk, line 88: IMO entities are better treated as immutable and modifying operationes moved to the repositories. In this case a method on the badge market repository that returns a new `BadgeBuyOffer` with decremented count and updated date.

TPP.Model/BadgeBuyOffer.cs
```diff
@@ -0,0 +1,90 @@
...
+            Id = id;
+            UserId = userId;
+            Species = species;
+            Form = form;
+            Source = source;
+            Shiny = shiny;
+            Price = price;
+            Amount = amount;
+            CreatedAt = createdAt;
+            WaitingSince = createdAt;
```
Felk, line 76: `WaitingSince = createdAt` even if just a little bit, but this looks like it puts a bit of logic into a data class. I think it's better for this to be passed in as well, and the caller can then decide to use the same date twice.

TPP.Model/BadgeBuyOffer.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using TPP.Common;
+
+namespace TPP.Model
+{
+    public class BadgeBuyOffer
```
Felk, line 11: looks like the entire thing could be a record

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+        {
+            List<BadgeBuyOffer> offers = await FindAllBuyOffers(userId, species, form, source, shiny);
+            if(offers.Count() < amount)
+                throw new ArgumentException(string.Format("Tried to cancel {0} offers but only {1} were found", amount, offers.Count));
+            
+            offers = offers.OrderByDescending(o => o.CreatedAt).ToList();
+            for(int i=0; i < amount; i++)
+            {
+                await BuyOfferCollection.FindOneAndDeleteAsync(o => o.Id == offers[i].Id);
+            }
```
Felk, line 134:
```suggestion
            foreach (BadgeBuyOffer offer in offers.OrderByDescending(o => o.CreatedAt).Take(amount))
            {
                await BuyOfferCollection.FindOneAndDeleteAsync(o => o.Id == offer.Id);
            }
```

Felk: same for `DeleteSellOffer` and possibly others

Felk, line 134: Could this possibly be batched? Feels like this should be 1 transaction. Same for DeleteSellOffer and possibly others

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+            badgesForSale.Reverse();
+            for(int i=0; i<amount; i++)
+            {
+                await _badgeRepo.SetBadgeSellPrice(badgesForSale[i], 0);
+            }
+        }
+
+        public async Task<ImmutableList<IBadgeMarketRepo.BadgeSale>> ResolveBuyOffers(PkmnSpecies species, bool? shiny)
+        {
+            List<Badge> badgesForSale = await FindAllBadgesForSale(null, species, null, null, shiny);
```
Felk, line 152: Please add the argument name when it isn't visible, for example `FindAllBadgesForSale(userId: null, species, form: null, source: null, shiny)`

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+            badgesForSale = badgesForSale.OrderByDescending(b => b.SellPrice).ThenBy(b => b.SellingSince).ToList();
+            List<IBadgeMarketRepo.BadgeSale> soldBadges = new List<IBadgeMarketRepo.BadgeSale>();
+
+            foreach (Badge badge in badgesForSale)
+            {
+                List<BadgeBuyOffer> buyOffers = await FindAllBuyOffers(null, species, null, null, shiny);
+                buyOffers = buyOffers.Where(o => o.Price >= badge.SellPrice).ToList();
+                buyOffers = buyOffers.OrderBy(o => o.WaitingSince).ToList();
```
Felk, line 160: Try to be a bit more LINQ-y. Don't unnecessarily call ToList on something when you could keep it an IEnumerable

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+
+                if(badge.SellPrice == null)
+                    throw new InvalidOperationException("Tried to sell a badge with no sell price");
```
Felk, line 163: how does that happen?

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// </summary>
+        private static List<Badge> SortBySpecialness(IEnumerable<Badge> toSort)
+        {
+            IEnumerable<Badge> duplicates;
+            IEnumerable<Badge> uniques = new List<Badge>();
+            HashSet<string?> formNames = new HashSet<string?>();
+            foreach (Badge b in toSort)
+            {
+                formNames.Add(b.Form);
+            }
```
Felk, line 214: LINQ be with you:
```suggestion
            HashSet<string?> formNames = toSort.Select(b => b.Form).ToHashSet();
```

Felk: the same applies to a bunch of places

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+                filter &= Builders<BadgeBuyOffer>.Filter.Eq(b => b.Form, form);
+            if (source != null)
+                filter &= Builders<BadgeBuyOffer>.Filter.Eq(b => b.Source, source);
+            if (shiny == true)
+                filter &= Builders<BadgeBuyOffer>.Filter.Eq(b => b.Shiny, true);
+            else
+                filter &= Builders<BadgeBuyOffer>.Filter.Ne(b => b.Shiny, true);
+
+            return await BuyOfferCollection.Find(filter).ToListAsync();
+        }
```
Felk, line 90: hmm not sure how I feel about this duplicate code. Especially since it is a bit finnicky in regards to the shiny filter requiring to filter for "!= true". If this is equal to the variant in BadgeRepo, can it be extracted as a utility method somewhere that builds a `FilterDefinition<T>`?

tests/TPP.Persistence.MongoDB.Tests/Repos/BadgeMarketRepoTest.cs
```diff
@@ -0,0 +1,218 @@
...
+            Instant after = before.PlusNanoseconds(69);
+
+            IBadgeMarketRepo badgeMarketRepo = CreateBadgeMarketRepo();
+
+            Badge badgeA = new Badge("A", userId, species, source, before, form, shiny);
+            Badge badgeB = new Badge("B", userId, species, source, after, form, shiny);
+            Badge badgeBForSale = new Badge("B", userId, species, source, after, form, shiny) { SellPrice = price };
+            List<Badge> notForSale = new List<Badge> { badgeA, badgeB };
+
+            _mockBadgeRepo.Setup(m => m.FindAllNotForSaleByCustom(userId, species, form, source, shiny)).Returns(Task.FromResult(notForSale));
```
Felk, line 86: iirc there's a `ReturnsAsync` that removes the need to wrap the result in `Task.FromResult(...)`

tests/TPP.Persistence.MongoDB.Tests/Repos/BadgeRepoTest.cs
```diff
@@ -18,12 +19,18 @@ public class BadgeRepoTest : MongoTestBase
         public BadgeRepo CreateBadgeRepo() =>
             new BadgeRepo(CreateTemporaryDatabase(), Mock.Of<IMongoBadgeLogRepo>(), Mock.Of<IClock>());
 
+        internal class MockClock : IClock
+        {
+            public Instant FixedCurrentInstant = Instant.FromUnixTimeSeconds(1234567890);
+            public Instant GetCurrentInstant() => FixedCurrentInstant;
+        }
+
```
Felk, line 27: I'm not sure why I used this class somewhere in the first place. I believe it is redundant because Moq mocks should work fine

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+                    
+                    soldBadges.Add(new IBadgeMarketRepo.BadgeSale(seller, buyer, badge, (long)badge.SellPrice));
+                    await _tokenBank.PerformTransactions(
+                        new Transaction<User>[]
+                        {
+                            new Transaction<User>(buyer, -1 * offer.Price, "BadgePurchase"),
+                            new Transaction<User>(seller, offer.Price, "BadgeSale")
+                        }
+                    );
+                    await _badgeRepo.TransferBadges(new List<Badge> { badge }.ToImmutableList(), buyer.Id, "BadgeSale", new Dictionary<string, object?>() { });
```
Felk, line 190: please put these transaction types in `TransactionType`

Felk (changes requested):

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
```diff
@@ -0,0 +1,249 @@
...
+                    break; //this badge has been sold, ignore the rest of the offers
+                }
+            }
+            return soldBadges.ToImmutableList();
+        }
+        /// <summary>
+        /// Sorts badges according to the priority in which they should be sold.
+        /// Current sorting rule: prioritize keeping 1 of each form, then sell newer badges first. Top priority badge will be selected for sale when fufilling offers.
+        /// </summary>
+        private static List<Badge> SortBySpecialness(IEnumerable<Badge> toSort)
```
Felk, line 206: I'm 95% sure with enough effort this can be expressed as a Comparer instead of an imperative sort method

Felk, line 206: a bunch of stuff, including this, probably needs more unit tests. Maybe you'll even have to make this sorter its own thing for that to be easier

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
Mogiiii: linq is magic to me, I'll read some documentation there.

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
Mogiiii: it absolutely shouldn't, but the null check here allows me to use badge.sellprice later on at 182 without the compiler complaining about a potentiall null value

TPP.Persistence.MongoDB/Repos/BadgeMarketRepo.cs
Mogiiii: amount isn't necessarily all of them.

Felk: note the `.Take(amount)`

Mogiiii: amazing what happens when you read.

PR #237, opened 2021-05-05
add multimonth gift subs support

Felk:


red031000 (commented):
bit iffy about using preview versions, as they can contain untested changes etc that might break other stuff

Felk:
> bit iffy about using preview versions, as they can contain untested changes etc that might break other stuff

you are correct, but TwitchLib in particular isn't exactly tested much anyway

red031000 (approved):

PR #238, opened 2021-05-07
avoid errors in async void methods

Felk:


chaos-lord (approved):
I can't test subs or whispers well, but it seems to both look and work fine to me

PR #240, opened 2021-05-09
static response commands

Felk:


aaaaaa123456789:
I'm not reading the code at the moment, but... are there any safeguards to prevent the dynamically-generated static commands from overriding the regular commands? It's only a matter of time until someone accidentally tries to create `!xp`.

Felk:
> I'm not reading the code at the moment, but... are there any safeguards to prevent the dynamically-generated static commands from overriding the regular commands? It's only a matter of time until someone accidentally tries to create `!xp`.

yes. It doesn't stop you from creating the command, but it will not overwrite existing ones and will log a warning instead.

aaaaaa123456789:
> > I'm not reading the code at the moment, but... are there any safeguards to prevent the dynamically-generated static commands from overriding the regular commands? It's only a matter of time until someone accidentally tries to create `!xp`.
> 
> yes. It doesn't stop you from creating the command, but it will not overwrite existing ones and will log a warning instead.

A proper response would be an error.

Felk:
> A proper response would be an error.

Yes, but that's not sufficient. Say a built-in command gets added after a static response command already exists. In that case you can't give an error at the time of creation. An error would be nice, but it would be in addition and not instead of the current handling. And I decided to spend the least effort on this.

red031000:
honestly I believe that the static commands and the config code removal should really be in two separate PRs

Felk:
It's a separate commit, isn't that good enough?

red031000:
I mean I guess, just means I need to review both at the same time

Felk:
ah, I see, that's why. I can do that if you want, though it's not a lot of code

TPP.Core/Setups.cs
```diff
@@ -159,5 +162,43 @@ public static (WebsocketBroadcastServer, OverlayConnection) SetUpOverlayServer(I
...
+        private static void SetUpDynamicCommands(
+            ILogger logger, CommandProcessor commandProcessor, IResponseCommandRepo responseCommandRepo)
+        {
+            IImmutableList<ResponseCommand> commands = responseCommandRepo.GetCommands().Result;
+
+            HashSet<string> dynamicallyInstalledCommands = new();
+            void InstallCommand(ResponseCommand command)
+            {
+                Command? existing = commandProcessor.FindCommand(command.Command);
+                if (existing != null)
```
red031000, line 175: this seems like a really weird way to do it, because the conflicting commands will be left in the repo and never cleaned up unless done manually, imo that should be done automatically

red031000 (changes requested):

TPP.Core/Setups.cs
Felk: I figured if some static command gets shadowed by a built-in command that it's appropriate that someone would need to take a look at it anyway

red031000:
ok, please fix the conflicts and I will approve

red031000 (approved):

PR #242, opened 2021-05-24
fix some log messages being dropped when aborting program

Felk:
the Abort handler for the ProcessExit and CancelKeyPress event was waiting for the mode to finish,
but failed to account for the fact that the process immediately terminates after that was done,
which in terms does not properly dispose stuff like the logger factory.
The latter is required for any remaining log messages to be flushed out.

### DIFF
--- a/TPP.Core/InputServer.cs
+++ b/TPP.Core/InputServer.cs
@@ -54,14 +54,14 @@ namespace TPP.Core
                     }
                     catch (HttpListenerException)
                     {
-                        _logger.LogInformation("input server listener was stopped.");
+                        _logger.LogDebug("input server listener was stopped");
                         return;
                     }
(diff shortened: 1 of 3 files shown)


chaos-lord (commented):
Nothing wrong from what i can see, but I don't know how to test this issue.

red031000 (changes requested):
line 53 on program.cs, please fix or suppress the warning, also instead of the explicit null check I think that a TryGetValue on the IDictionary would be better, purely as a style choice (I know this isn't directly related but it came up as a warning on this PR so I'd prefer to have it fixed if possible). I also don't appear to be able to comment on that line on mobile, seems github needs to fix their shit

red031000 (approved):

PR #262, opened 2021-06-07
prevent two dualcore instances running simultaneously using exclusivity token

Felk:


TPP.Core/DatabaseLock.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace TPP.Core
+{
+    internal class DatabaseLockEntry
+    {
+        public const string KeyValueId = "database_lock";
+        [BsonId]
+        public string Id { get; init; } = KeyValueId;
+        [BsonElement("refreshed_at")]
+        public Instant RefreshedAt { get; init; }
+    }
```
Felk, line 18: hmm... mongodb specific stuff within the core project, I'm not sure I like that

Felk:
TODO after merge: enable the service to launch on boot on the i5-emu

red031000 (approved):

PR #280, opened 2021-09-20
commands cooldowns using load factors

Felk:


Dnantz (changes requested):
Is there a way to specify some commands to ignore/add less load factor (E.G. run inputs that are expected to be spammed) and conversely heavy commands (E.G. that do lots of database operation) would cost more load factor,  perhaps as a property of the command itsself. If this ends up adding more complexity than benefit feel free to disregard.

PR #292, opened 2021-11-05
strip trailing unicode tags in incoming messages

Felk:


Felk:
pls also merge https://github.com/TwitchPlaysPokemon/tpp/pull/509

PR #293, opened 2021-11-12
prepare for .NET 6

Felk:
This does not set the target framework to 6 yet, but already runs the workflows with .NET 6 and claims .NET 6 is required in the READMEs. It's a non-breaking preparation for switching over to .NET 6


Felk:
Oh, I didn't know .NET 6 runtime refuses to run apps targeted towards .NET 5

PR #307, opened 2022-03-22
add dualmode support to runmode

Felk:


Felk:
also needs https://github.com/TwitchPlaysPokemon/tpp/pull/512

PR #314, opened 2022-04-03
run features: emblems, input counter, input muting

Felk:


Felk:
Regarding input muting: muting can be enabled at startup by configuring `MuteInputsAtStartup` in the runmode config to true. At runtime, either hit `http://127.0.0.1:5010/start_run` or do `!unmuteinputs`. You can also enable the muting by hitting `http://127.0.0.1:5010/stop_run` or doing `!muteinputs`.

PR #316, opened 2022-04-05
add some more features for sided inputs

Felk:
- [X] Allow `l`, `left`, `r`, `right` as side and `:`, `.` `@`, `#` as delimiter.
- [x] Allow choosing side with `!left` or `!right`, letting people omit the side prefix afterwards.
- [x] Add config to disallow directed inputs and force people to pick a side instead.
- [ ] Add configurable throttle for how often people can switch sides. Default is no throttle.

### DIFF
--- a/TPP.Core/Modes/Runmode.cs
+++ b/TPP.Core/Modes/Runmode.cs
@@ -27,7 +27,6 @@ namespace TPP.Core.Modes
         private readonly WebsocketBroadcastServer _broadcastServer;
         private AnarchyInputFeed _anarchyInputFeed;
         private readonly OverlayConnection _overlayConnection;
-        private readonly InputBufferQueue<QueuedInput> _inputBufferQueue;
 
         private readonly StopToken _stopToken;
         private readonly MuteInputsToken _muteInputsToken;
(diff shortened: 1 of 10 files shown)


TPP.Core/Modes/Runmode.cs
```diff
@@ -98,12 +97,38 @@ private AnarchyInputFeed CreateInputFeedFromConfig(InputConfig config)
...
+        private async Task ProcessPotentialSidedInputs(User user, InputSequence inputSequence)
+        {
+            foreach (InputSet inputSet in inputSequence.InputSets)
+            {
+                if (inputSet.Inputs.FirstOrDefault(i => i is SideInput) is SideInput { Side: null } sideInput)
+                {
+                    string? side = await _inputSidePicksRepo.GetSide(user.Id);
+                    if (side == null)
+                    {
+                        // TODO auto-assign or flip flop or deny input or something?
```
Felk, line 121: TODO

TPP.Core/Setups.cs
```diff
@@ -86,6 +86,7 @@ public static ArgsParser SetUpArgsParser(IUserRepo userRepo, PokedexData pokedex
                 ).Commands,
                 new PollCommands(databases.PollRepo).Commands,
                 new ManagePollCommands(databases.PollRepo).Commands,
+                new InputtingCommands(databases.InputSidePicksRepo).Commands,
```
Felk, line 89: This should probably just be added during runmode

TPP.Core/Modes/Runmode.cs
m4-used-rollout: Flip flop

PR #319, opened 2022-04-15
add moderation commands

Felk:
- [x] `!timeout` and `!untimeout`
- [x] `!ban` and `!unban`
- [x] `!checkban` and `!checktimeout`
- [ ] ~~during matchmode: clear move selection after punitive moderator actions~~ matchmode doesn't do betting yet anyway
- [ ] ~~deputy notices. The deputy system itself isn't ported yet either~~ I'm not gonna bother with that yet. If it becomes urgent I'll add it in a separate merge request
- [ ] ~~check whether modbot should log to banlog and timeoutlog too. Feels like it should and I just misread old core's code.~~ Nope, old core only logs manual bans and timeouts in "banlog" and "timeoutlog". The automatedones are logged in "modlog".


TPP.Core/Moderation/ModerationService.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        ModerationActionPerformed?.Invoke(this, new ModerationActionPerformedEventArgs(
+            issuerUser, targetUser, isIssuing ? ModerationActionType.Timeout : ModerationActionType.Untimeout));
+
+        // TODO is this obsolete?
+        await _messageSender.SendWhisper(targetUser,
+            $"You were {(isIssuing ? "timed out" : "untimed out")} " +
+            $"by {issuerUser.Name} at {now}. Reason: {reason}");
+
+        return TimeoutResult.Ok;
```
Felk, line 118: Users already get timeout and ban reasons displayed in chat. Is this still needed? Only advantage I see is that they also see any potential _unban_ or _untimeout_ reasons.

TPP.Core/Modes/ModeBase.cs
```diff
@@ -133,6 +134,11 @@ private async Task ProcessIncomingMessage(IChat chat, Message message)
             {
                 return;
             }
+            if (message.User.Banned || message.User.TimeoutExpiration > now)
+            {
+                _logger.LogDebug("Skipping message from {User} because user is banned or timed out", message.User);
+                return;
+            }
```
Felk, line 141: fyi, this implies that these messages also are not forwarded to old core. So any banned or timed out user is cut off (only affects whispers since they wouldn't be able to write in chat anyway)

TPP.Core/Moderation/ModerationService.cs
Felk: I assumed it's unnecessary and removed it

PR #320, opened 2022-04-18
more dualgb related features

Felk:


Felk:
try this config to disable all features: `"DisabledFeatures": ["Badges", "Cosmetics", "Currencies", "Polls"],`

Felk:
@m4-used-rollout I just spitballed what features need to be disabled. Please take a look. There are some open issues to address:
- [ ] How should incoming subs be handled? Right now it handles them normally. Possible solution: Handle normally but don't emit any overlay events or whispers.
- [ ] How should incoming donations be handled? Right now new core does not handle any donations. That might be a problem.
- [ ] Should the user object fields of the features that are disabled be removed from overlay events? Or is it enough to trust the event overlay to just ignore those? And if yes, is that also a good long-term solution?

m4-used-rollout:
- Handle normally is fine. If people sub and wonder about tokens, we can tell them that they can be used during PBR
- Doesn't the core have to poll for donations? If someone donates while the old core is down, will the donation be processed when the old core comes back up later?
- No, don't worry about limiting the data sent to the overlay. The event input feed should decide what to show, not the core itself.

Felk:
> Handle normally is fine.

Ok good, nothing to do then.

> If someone donates while the old core is down, will the donation be processed when the old core comes back up later?

Hmm, looks like that's what the `refresh_all_tips_worker` does, yes.

> The event input feed should decide what to show, not the core itself.

Good, nothing to do then either.

TPP.Core/Configuration/RunmodeConfig.cs
```diff
@@ -21,5 +22,9 @@ public class RunmodeConfig : ConfigBase, IRootConfig
                      "Inputs can be unmuted at runtime using !unmuteinputs or calling /start_run on the inputserver. " +
                      "Inputs can be muted at runtime using !muteinputs or calling /stop_run on the inputserver. ")]
         public bool MuteInputsAtStartup = false;
+
+        [Description("If not null, this amount of time must pass before a player can switch sides again." +
+                     "Only relevant for dual-sided input profiles.")]
+        public TimeSpan? SwitchSidesCooldown { get; init; } = null;
```
Felk, line 28: hmm... is this required to be updateable at runtime?

Felk: it is now

PR #321, opened 2022-04-24
add democracy

Felk:


TPP.Core/Overlay/Events/RunInputEvents.cs
```diff
@@ -66,11 +67,81 @@ public sealed class AnarchyInputStop : IOverlayEvent
...
+    public sealed class DemocracySequenceStart : IOverlayEvent
+    {
+        public string OverlayEventType => "democracy_sequence_start";
+
+        [DataMember(Name = "button_sequence")] public List<List<string>> Sequence { get; init; }
+        public DemocracySequenceStart(InputSequence input)
+        {
+            Sequence = input.InputSets
+                .Select(set => set.Inputs
+                    .Select(i => i.DisplayedText).ToList()).ToList();
```
Felk, line 144: The old overlay relies on the input text to be regex-parseable to be recognized as a touch input and therefore displayed during execution, so this does not work for touchscreen aliases. Might be worth rethinking the wire format.
Note that in the old overlay the wire format for button sequences is shared across many places, and in particular with the `anarchy_input_start` event in regards to the touch detection logic (see `check_button_set_for_touchscreen` in old overlay `touchscreen.js`).

PR #337, opened 2022-07-24
Attempt to fix undetected twitch chat disconnects

ferraro2:


Felk:
guess this didn't fix it then :(

ferraro2:
I don't remember seeing this before where the old core stays connected to twitch, can send messages, but doesn't receive them.  A problem like that might be an issue on twitch's end.

The last two new core disconnects happened during a run and after an internet outage, so it's a little hard to tell if this is the same issue. It might be though

PR #392, opened 2024-02-06
refactor chat connections and allow dynamically joining chats

Felk:
- [x] entire co-stream-inputs feature can be enabled using `CoStreamInputsEnabled` (inside twitch connection config) (default false)
- [x] !join and !leave are implemented
- [x] !join only works when the user is live (configurable via `CoStreamInputsOnlyLive`, default true)
- [x] if `CoStreamInputsOnlyLive` is set to true, automatically leaves channgels that are no longer live
- [x] channels joined this way are only monitored for mode-specific tasks, for runmode this means only inputs. No modbot or commands.
- [x] only a message after !join and a message after !leave are posted in the target chats, otherwise the bot is quiet there
- [x] inputs from TPP-banned users are ignored from all channels
- [x] anarchy inputs got a new "channel" field that is null for tpp and populated with a channel name for foreign inputs
- [x] !join and !leave are persisted across core restarts
- [x] It ignores co-stream inputs while the main channel is in emote-only mode
- [x] channel icon URLs for the channels provided
- [ ] (?) !leaveall command
  - could just set `CoStreamInputsEnabled` to false after the event and it stops joining other channels
- [x] if TPP’s channel is in slow mode, it should not accept inputs at higher speed from anywhere else
- [x] !left and !right still work


TPP.Core/Chat/TwitchChatQueue.cs
```diff
@@ -97,47 +95,57 @@ private async Task ProcessOne()
...
         private async Task Send(OutgoingMessage message)
         {
+            TwitchAPI twitchApi = await _twitchApiProvider.Get();
             if (message is OutgoingMessage.Chat chat)
-                await _twitchClient.SendMessageAsync(chat.Channel, chat.Message);
+                await twitchApi.Helix.Chat.SendChatMessage(chat.ChannelId, _senderUserId, chat.Message);
             else if (message is OutgoingMessage.Reply reply)
-                await _twitchClient.SendReplyAsync(reply.Channel, replyToId: reply.ReplyToId, message: reply.Message);
+                // TODO https://github.com/TwitchLib/TwitchLib.Api/pull/386
+                await twitchApi.Helix.Chat.SendChatMessage(reply.ChannelId, _senderUserId, reply.Message /*, replyParentMessageId: reply.ReplyToId*/);
```
Felk, line 103: Need to fix this bit if/when https://github.com/TwitchLib/TwitchLib.Api/pull/386 is merged and available on nuget via a prerelease

Felk: I've disabled twitch-native replies for now in 620b440f1c0016218159e7093fbf315bcc4b820c until we can use the API for them again. So we need to also revert 620b440f1c0016218159e7093fbf315bcc4b820c once this is fixed

PR #413, opened 2024-06-03
Add configurable additional input map keys when queue depletes or refills

Felk:


Felk:
This works as follows:

- Configure the input config with input map keys of your liking, e.g.:
  ```json
  "InputConfig": {
    // ...
    "QueueEmptyInputMapKey": "Pause",
    "QueueNoLongerEmptyInputMapKey": "Unpause"
  },
  ```
  or just `"Start"` for both if the input controller itself doesn't get enhanced with any semantic interpretations
- The input map endpoint can now be observed to behave as follows:
  ```
  // start with empty queue
  >> GET /gbmode_input_request_bizhawk
  << {}
  // input "A"
  // input "B"
  >> GET /gbmode_input_request_bizhawk
  << {"A":true,"Held_Frames":16,"Sleep_Frames":84,"Input_Id":1717433953866,"Unpause":true}
  >> GET /gbmode_input_request_bizhawk
  << {"A":true,"Held_Frames":16,"Sleep_Frames":84,"Input_Id":1717433953866}
  >> GET /gbmode_input_complete
  >> GET /gbmode_input_request_bizhawk
  << {"B":true,"Held_Frames":16,"Sleep_Frames":84,"Input_Id":1717434101957}
  >> GET /gbmode_input_complete
  >> GET /gbmode_input_request_bizhawk
  << {"Pause":true}
  >> GET /gbmode_input_request_bizhawk
  << {}
  ```
- These additional input map keys are not actual inputs to the rest of the system.
  They only get added to the input map and do not e.g. appear as inputs on the overlay or are logged anywhere.

The current implementation has the following limitations:
- Currently, these special inputs only get added once per poll. In the case that the input controller actually polls `/gbmode_input_request_bizhawk` twice because it dropped once or something, the Pause/Unpause input will not be repeated.  
  If this is a problem, it can be changed, but I wanted to get an opinion on this first.
- It arbitrarily treats a fresh boot as an empty queue, so the "QueueNoLongerEmptyInputMapKey" will get added for the first input after a restart. This can be arbitrarily flipped, but nothing more elaborate currently.

m4-used-rollout:
- Repeated input requests without an input complete should always return identical input maps, ideally. Please see about making the queue end/queue restart state persist until the input complete endpoint is requested.
- For the upcoming Tetris event, there is no possibility that the input puller will be able to understand Pause and Unpause. We’ll have to make do with sending Start for both.
- To this end, we should probably disallow users from being able to input the pause/unpause buttons. Perhaps as a config setting?

m4-used-rollout:
Also, this behavior is controlled entirely by the presence of QueueEmptyInputMapKey or QueueNoLongerEmptyInputMapKey, right? These don’t have default values?

Felk:
> Also, this behavior is controlled entirely by the presence of QueueEmptyInputMapKey or QueueNoLongerEmptyInputMapKey, right? These don’t have default values?

Yes. Their defaults are `null`, meaning the feature is disabled

Felk:
> Repeated input requests without an input complete should always return identical input maps, ideally. Please see about making the queue end/queue restart state persist until the input complete endpoint is requested.

fixed in 3ee4c7c5bcc15911ccc0e5411025fa7470c628ea

> For the upcoming Tetris event, there is no possibility that the input puller will be able to understand Pause and Unpause. We’ll have to make do with sending Start for both.

That's okay, just configure this then:  
```json
"InputConfig": {
  // ...
  "QueueEmptyInputMapKey": "Start",
  "QueueNoLongerEmptyInputMapKey": "Start"
},
```

> To this end, we should probably disallow users from being able to input the pause/unpause buttons. Perhaps as a config setting?

Should I add safeguards for this or is it acceptable to forbid this "by policy" by manually just not including whatever button was configured in the button profile?

m4-used-rollout:
> For your use-case having a dedicated button profile for Tetris that just doesn't have "Start"

That can work for now.

PR #418, opened 2024-07-06
Update Emblems.cs

Ravendwyr:
I'm probably jumping the gun here


TPP.Common/Emblems.cs
```diff
@@ -96,6 +96,9 @@ public static class Emblems
             [87] = "Super Gauntlet Colosseum",
             [88] = "Super Gauntlet XD",
             [89] = "Super Gauntlet Black 2/White 2",
+            [90] = "Super Gauntlet Sword/Shield",
+            [91] = "Super Gauntlet Legends: Arceus",
+            [92] = "Super Gauntlet Scarlet/Violet",
```
m4-used-rollout, line 101: 91 and 92 are backwards

Ravendwyr:
Like this?

Issue #401, opened 2024-03-17
Command Aliases


m4-used-rollout:
Much like how we can add static responses through the !responses command, it would be great if we could add aliases through an !aliases command.

For instance, during dual runs we have `!left` and `!right`. It would be great if a moderator or operator could also add `!gold` and `!silver` (for example) that map to `!left` and `!right`. Basically, if a user used an aliased command `!gold`, the command parser would replace it with `!left` and run it back through the parser again.

the aliased commands should probably translate to an entire string to allow for parameters to be hardcoded, but then also pass the rest of the command string so extra parameters can be added. For instance, we could have a `!bestbadges` command that would map to `!dex complement-from-dupes` so a user could say `!bestbadges m4_used_rollout` to execute `!dex complement-from-dupes m4_used_rollout`

Felk:
It's now fully addressed by #423

PR #423, opened 2024-08-01
add dynamic command aliases

Felk:
Partially fixes https://github.com/TwitchPlaysPokemon/tpp-core/issues/401

no hard-coded arguments as-is

### DIFF
--- a/TPP.Core/Chat/ICommandHandler.cs
+++ b/TPP.Core/Chat/ICommandHandler.cs
@@ -1,0 +1,8 @@
+using System.Threading.Tasks;
+
+namespace TPP.Core.Chat;
+
+public interface ICommandHandler
+{
+    public Task ProcessIncomingMessage(IChat chat, Message message);
+}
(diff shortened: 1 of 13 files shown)


Felk:
Now it also has hard-coded arguments:

```
>>> !aliases add felk badges @felkcraft
Command alias 'felk' now redirects to: badges @felkcraft
>>> !aliases list
These command aliases currently exist: felk -> badges @felkcraft
>>> !felk
FelkCraft has no badges.
>>> !aliases remove felk
Command alias 'felk' was removed.
>>> !aliases list
There currently are no command aliases.
>>> !felk
unknown command 'felk'
```

Felk:
waiting for feedback, but will eventually merge anyway if noone objects

m4-used-rollout:
Does it also append any parameters given to the aliased command?
> For instance, we could have a !bestbadges command that would map to !dex complement-from-dupes so a user could say !bestbadges m4_used_rollout to execute !dex complement-from-dupes m4_used_rollout

Felk:
Yes!

m4-used-rollout:
Can it be used to trigger commands in old core?

Felk:
Oh, hmm... Probably not but I don't know, haven't actually tried that

Felk:
It doesn't work right now, and it is surprisingly complicated. How important is that?

Felk:
89ee4819c3376ad6451265619cc21463262f6363 looks like it does the job, even if it's quite hacky. Does that look good so far?

m4-used-rollout:
If it works it’s fine with me

PR #436, opened 2024-10-09
Refactor tpp-core directory structure

feselene:
The main purpose of this PR is to move all of the source files into a src folder separate from the root directory where the other project files are being kept. Refactor tpp-core repo to use PascalCase C# naming conventions. 

Automated checks are passing and `dotnet test` tests are passing as well. 

image
image

image


Felk:
First off, hello there 👋 
Thanks for spending some time trying to contribute.

Now regarding your changes, I'm afraid I'll have to apply some brakes here.
As I understand it, this is basically just for the sake of "Best Practices", right? This is a big change, as it causes conflicts with all open branches and forks. For this reason I'd like to have some tangible benefits for this change.

The current naming convention doesn't seem that unconventional. For example, TwichLib also prepends all their packaged with "TwitchLib.". Probably because they want their imports to also start with that name to avoid namespace confusion.

feselene:
Hello @Felk 

I understand that this change will cause conflicts with all open branches and forks.

I should be able to reimplement all of the open branches and forks within 72 hours. I have already reimplemented the dynamic_command_aliases PR here. https://github.com/TwitchPlaysPokemon/tpp-core/pull/439

The changes are included in commit [a238f6c](https://github.com/TwitchPlaysPokemon/tpp-core/pull/439/commits/a238f6c259a73b9a0b1c665a839365c31ff2e6d3)

feselene:
> First off, hello there 👋 Thanks for spending some time trying to contribute.
> 
> Now regarding your changes, I'm afraid I'll have to apply some brakes here. As I understand it, this is basically just for the sake of "Best Practices", right? This is a big change, as it causes conflicts with all open branches and forks. For this reason I'd like to have some tangible benefits for this change.
> 
> The current naming convention doesn't seem that unconventional. For example, TwichLib also prepends all their packaged with "TwitchLib.". Probably because they want their imports to also start with that name to avoid namespace confusion.

I feel like appending TPP. in front of most of the package names is unnecessary if we just put all of the source files into a src/ folder. It's good practice to keep all of the source files in a separate src folder rather than in the root directory with all of the other project files.

Felk (commented):
Sorry, as a first time contributor I'd prefer it if you would start off with something less intrusive. Let's try to build up some trust first.

Regarding tangible benefits, "it's good practice" is not a benefit. Good practices emerge _from_ tangible benefits.  
How is it better and not just different?

feselene:
> Sorry, as a first time contributor I'd prefer it if you would start off with something less intrusive. Let's try to build up some trust first.
> 
> Regarding tangible benefits, "it's good practice" is not a benefit. Good practices emerge _from_ tangible benefits. How is it better and not just different?

I'm fine with building up trust first. I'll implement the entire twitchplayspokemon website first.

Felk:
> I'm fine with building up trust first.

Thanks. I'll close this one then.

> I'll implement the entire twitchplayspokemon website first.

That's promising to hear 😃 let's continue in that respective pull request

PR #466, opened 2025-04-04
revert badge rarity calculation to go back to when gen2 was added

Felk:
Branch just for trying out how badge rarities behave, don't merge.


Felk:
was never needed after all

